PR #9, opened 2016-04-07
Setup phpunit + travis

stefk:
Also fixes result bundle tests.


ngodfraind:
Should I squash these commits ? We can do if from github easily without cli.

stefk:
I'd like to keep at least two separate commits (travis + result bundle). I'll rebase myself after figuring out what's wrong on travis.

stefk:
@ngodfraind, it works. I squashed what I could but the merge in the middle makes it hard to go further. I think it's best to merge this as is, I'll be more careful the next time.

PR #120, opened 2016-04-08
Fixing twig serializer filter

ngodfraind:


ngodfraind:
LGTM

stefk:
It's hard to review without any description, doc or calling code. From what I can make out there's no fix, you're just allowing `api_serialize` to serialize other groups than `api`, which is confusing.

ngodfraind:
Wich is exactly what it's supposed to do.

{{ myEntities|api_serialize('mySerializationGroup) }}

stefk:
I think what you're looking for is simply a general purpose `serialize` function, not a function which something different than what its name implies...

ngodfraind:
I don't know where it's used (it's not used in the distribution bundle but it could be used elsewhere - maybe Formalibre/BulletinBundle or Claroline/TranslatorBundle or somewhere else. I don't know.

I know it's going to be used in the new FlashCardBundle (or at least something similar).

For instance, in your ResultBundle, you do this.

```
        window.resultId = {{ _resource.getId() }}
        window.resultMax = {{ _resource.getTotal() }}
        window.resultMarks = {{ marks|json_encode|raw }}
        window.workspaceUsers = {{ users|json_encode|raw }}
        window.isReadOnly = {{ canEdit ? 'false' : 'true' }}
```


It works beause you're returning arrays. Sometimes, we return entities and we want to the same.
```
        window.resultId = {{ _resource.getId() }}
        window.resultMax = {{ _resource.getTotal() }}
        window.resultMarks = {{ marks|api_serialize('api_mark')|raw }}
        window.workspaceUsers = {{ user|api_serialize('api_user')|raw }}
        window.isReadOnly = {{ canEdit ? 'false' : 'true' }}
```

This is a tiny change (wich is more a bugfix than anything else because it's what was intended).

Should we make such exhaustive justification for everything ? Because It's going to take a long while in t.

stefk:
You don't need to give an exhaustive justification like that. All I'm saying is that it's hard to declare something good when you have *zero* information on it. On the other hand, we're all moving to an api-based architecture, so even if it's a tiny change, it's useful to everybody and it's important to understand it and get it right.

@claroline/maintainers, any thoughts on this process?

`serialize` solves the problem IMO but if you're not sure about `api_serialize` usage, maybe you should also keep it as a shortcut for `serialize('api')` to avoid breaking existing code.

stefk:
Proposal:

```php

/**
 * Serializes data to JSON, optionally filtering on a serialization group.
 *
 * @param mixed  $data
 * @param string $group
 */
public function serialize($data, $group = null)
{
    $context = new SerializationContext();
    
    if ($groups) {
        $context->setGroups($group);
    }

    return $this->container->get('serializer')->serialize($data, 'json', $context);
}

/**
 * Serializes data to JSON using the "api" serialization group.
 *
 * @param mixed $data
 */
public function apiSerialize($data) 
{
    return $this->serialize($data, 'api');
}

```

ngodfraind:
I'm trying to do some bugfixes on the path bundle and I'll do that after.

2016-04-11 11:06 GMT+02:00 Stéphane Klein <notifications@github.com>:

> Proposal:
>
> /** * Serializes data to JSON, optionally filtering on a serialization group. * * @param mixed  $data * @param string $group */public function serialize($data, $group = null){    $context = new SerializationContext();    if ($groups) {        $context->setGroups($group);    }    return $this->container->get('serializer')->serialize($data, 'json', $context);}/** * Serializes data to JSON using the "api" serialization group. * * @param mixed $data */public function apiSerialize($data) {    return $this->serialize($data, 'api');}
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/claroline/Distribution/pull/120#issuecomment-208242134>
>



-- 
Godfraind Nicolas
5, rue de la Carrière
6592 Monceau-Imbrechies
GSM: 0472 94 80 99

stefk:
Will probably clash with built-in filter http://jmsyst.com/bundles/JMSSerializerBundle#usage...

stefk:
`json_serialize`, to match [json_encode](http://twig.sensiolabs.org/doc/filters/json_encode.html#json-encode)?

ngodfraind:
Agree

2016-04-11 11:16 GMT+02:00 Stéphane Klein <notifications@github.com>:

> json_serialize, to match json_encode
> <http://twig.sensiolabs.org/doc/filters/json_encode.html#json-encode>?
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/claroline/Distribution/pull/120#issuecomment-208245084>
>



-- 
Godfraind Nicolas
5, rue de la Carrière
6592 Monceau-Imbrechies
GSM: 0472 94 80 99

ngodfraind:
LGTM

stefk:
LGTM

PR #123, opened 2016-04-11
q&d patch for condition update

ngodfraind:
I don't think it should be merged, but I had to do this on a production instance to make it work.

Fatal error: arg 2 passed to StepConditionManager::edit() must be an instance of StepCondition, null given.
Called in StepManager line 140.


stefk:
ping @claroline/innova

Elorfin:
LGTM.

For issue about Path conditions, I think we have to ping @stmolivier too.

LaurentGruber:
LGTM

ngodfraind:
LGTM

PR #125, opened 2016-04-11
Bundle activation

ngodfraind:
This is a work in progress but it'll allow you the see what's already done.


ngodfraind:
It's not over yet (it still needs translations and a few warning / insights for the users)
- the bundle.ini file can be changed with the gui
- most obvious request are updated
- the version and package of the plugins are store in database

In order to enable/disable plugins programmatically:

$plugin = $pluginRepo->findBy(...);
$this->getContainer()->get('claroline.manager.bundle_manager')->enable($plugin);

I'll add an optional require.json file at the root of every plugin wich will trigger then disable function on installation if needed and warn that the activation is not possible with the gui.

The annoying/dirty part is editing requests as it requires the BundleManager::getEnabled() method to retrieve the list of enabled bundle. There is no flag is the Plugin entity to avoid side effects (enabled in plugins but disabled in bundles.ini for instance). It's annoying for us but it makes a better user experience.

main/core/Command/PlatformInstallCommand.php
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        /*
+         * Set the app/config directory in the installation state.
+         * - No bundles.bup.ini
+         * - Empty previous-installed.json
+         */
+		$kernel = $this->getContainer()->get('kernel');
```
stefk: There's a lot of spacing issues in this PR (hard tabs vs soft tabs, indentation, etc.).

main/core/Controller/API/Admin/PluginController.php
```diff
@@ -0,0 +1,84 @@
...
+use FOS\RestBundle\Controller\Annotations\NamePrefix;
+use FOS\RestBundle\Controller\Annotations\Post;
+use FOS\RestBundle\Controller\Annotations\Get;
+use FOS\RestBundle\Controller\Annotations\Patch;
+use Claroline\CoreBundle\Entity\Plugin;
+
+/**
+ * @NamePrefix("api_")
+ */
+class PluginController extends FOSRestController
```
stefk: Is there a security check somewhere for this controller?

main/core/Entity/Plugin.php
```diff
@@ -22,35 +24,66 @@
...
      * @ORM\Column(name="has_options", type="boolean")
+     * @Groups({"api_plugin"})
      */
     protected $hasOptions = false;
 
+    /**
+     * @ORM\Column(name="version")
+     * @Groups({"api_plugin"})
+     */
+    protected $version;
```
stefk: Looks like a step back. Composer should be the only source of truth for versionning and package info IMO.

main/core/Library/Utilities/FileSystem.php
```diff
@@ -31,8 +31,10 @@ public function rmDirContent($path, $recursive = false)
         $iterator = new \DirectoryIterator($path);
 
         foreach ($iterator as $el) {
```
stefk: Not strictly related to this PR, but this whole class seems useless considering there's a `Filesystem` component doing all that (and more) perfectly.

ngodfraind: This class extends the symfony2 Filesystem to add recursive directories suppression.

main/core/Command/PlatformInstallCommand.php
ngodfraind: I noticed. I need to configure my ide.

main/core/Library/Utilities/FileSystem.php
stefk: The original `remove` is already recursive. If you want to remove only the content you can pass it an iterator.

main/core/Repository/AdministrationToolRepository.php
```diff
@@ -12,9 +12,31 @@
...
 
 use Doctrine\ORM\EntityRepository;
+use JMS\DiExtraBundle\Annotation as DI;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 
-class AdministrationToolRepository extends EntityRepository
+class AdministrationToolRepository extends EntityRepository implements ContainerAwareInterface
 {
+    public function setContainer(ContainerInterface $container = null)
```
stefk: I don't see this method called anywhere (same for 5 other occurrences in repositories). Plus 2 problems:
- unrelated side effect in a setter is bad
- repository methods are broken if setter isn't called
- `bundles` property isn't declared or initialized properly

ngodfraind: It implements ContainerAware interface.

It works. Unless you have a better solution for retrieving the bundles.ini list easily...

stefk: Ok if it's an automatic injection. But at least smth like that:

```php
private $container;
private $bundles;

public function setContainer(ContainerInterface $container)
{
    $this->container = $container;
}

private function getEnabledBundles() 
{
    if (!$this->bundles) {
        $this->bundles = $this->container->get('claroline.manager.bundle_manager')->getEnabled(true);
    }

    return $this->bundles;
}
```

ngodfraind:
I'm going to change how the installation work here because it's related to plugin enable/disable thing.

ngodfraind:
The installation/update is now refactored.
It'll still need some changes because we want to install/update bundles regardless of their enable/disable status.
Also, the bundles.ini file generation should not override everything like now.

main/core/Migrations/pdo_mysql/Version20160413112256.php
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * Generation date: 2016/04/13 11:22:58
+ */
+class Version20160413112256 extends AbstractMigration
+{
+    public function up(Schema $schema)
+    {
+        $this->addSql("
+            ALTER TABLE claro_plugin 
+            ADD description VARCHAR(255) DEFAULT NULL
```
stefk: Question: where does the description come from?

ngodfraind: Nowhere yet, but I think we could add some description and usage for most plugins we have. In that branch, it'll look for a file named 'DESCRIPTION.md' and put it in the database. The database field may be wrong atm and I'm going to remove it anyway.

I'll display it directly in the plugin interface.

stefk: If it's a short, non-translated description, composer's [description](https://getcomposer.org/doc/04-schema.md#description) could be enough. Otherwise a convention like you suggest (md/text file(s)) is good. Either way the db field is not needed IMO.

stefk:
(BTW still some space/tab issues, also in html/js files).

ngodfraind:
There is no composer description for each plugins of a meta package, wich
is the reason why did it.

(I know about the html/js file, I did not configure my editor yet and I
didn't write anything in those files since then. They need some updates and
I'll fix the tabbing at that moment).

2016-04-13 19:00 GMT+02:00 Stéphane Klein <notifications@github.com>:

> (BTW still some space/tab issues, also in html/js files).
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/claroline/Distribution/pull/125#issuecomment-209546251>
>



-- 
Godfraind Nicolas
5, rue de la Carrière
6592 Monceau-Imbrechies
GSM: 0472 94 80 99

ngodfraind:
+ Basic php extension check for metapackages plugin.

ngodfraind:
- Maintenance environment. This environment loads every plugin no matter what. 

@stefk 
Install/update operations should be done regardless of plugin's activation status in order to avoid conflicts during update later on.

Currently, I'm forcing the installation/update using the "--env=maintenance" parameter, but it could cause issues (for instance, there is no way to run the command in the test environment for now but I could just change my condition - not every plugin will be loaded that way). 

Also, I can't just switch the kernel environment and reboot it during this command (it causes cache issue, and it doesn't really feel good anyway).

stefk:
[In response to https://github.com/claroline/Distribution/pull/131#issuecomment-209914933:]

It could be a dead end in the long run. Consider these use cases:

- a plugin needs a lib, but it must be in a particular version range
- a plugin needs a lib, but it can fallback to another one if missing
- a plugin needs several conditions to be met, others than lib/ext/plugin (e.g. a particular service running on a given port)

The core cannot handle all this. The solution IMO is to let plugins decide if the conditions are met and report back to the core. Using php interfaces is also probably the best option.

> It will need to support plugin requirements aswell (some plugins depends on each others).

Agreed. Again, extending one of the bundle interfaces could be a way to do it.

ngodfraind:
Ok, therefore, let's add 2 functions in the PluginInterface:

getRequiredPhpExtension() {}
getRequiredPlugins() {}

and drop the require.json file.
Extensions and Plugins are easy to handle for the core bundle. I won't use any version for the lib for now. Plugins will be able to handle it much better if they want to but it's an easy lazy way to deal with most use cases.

By the way, what do you think of the 'maintenance' environement and only allowing update operations from it ? We really want each plugins to be updated/installed no matter what imo.

stefk:
> Currently, I'm forcing the installation/update using the "--env=maintenance" parameter, but it could cause issues (for instance, there is no way to run the command in the test environment for now but I could just change my condition - not every plugin will be loaded that way).

Using a separate environment is risky because not every bundle will know it, especially third-party ones. It could lead to unexpected behaviours. Plus switching environmnents during execution is a real pain (`SwitchKernel` never really worked well; its method have been used a few weeks max and it just stayed there for compatibilty reasons).

ngodfraind:
I noticed switching environment didn't really worked here. The thing is that we need every plugin loaded to perform updates and it's the most straightforward way to achieve it.

stefk:
>Ok, therefore, let's add 2 functions in the PluginInterface:
>
>getRequiredPhpExtension() {}
>getRequiredPlugins() {}
>
>and drop the require.json file.

+1. And in the future, maybe something like `getFailedRequirements()` to make additional/custom checks. But are we sure it's only at the plugin level we need this?

stefk:
> I noticed switching environment didn't really worked here. The thing is that we need every plugin loaded to perform updates and it's the most straightforward way to achieve it.

You use the true/false check in bundles.ini? If so, backing up that file and replacing it temporarily with a copy with all flags set to true may do the trick.

ngodfraind:
It could be done during composer update but it wouldn't allow to perform a clean install after an update (unless you change it manually).
While it's not a problem for production, it's very annoying in dev.

Otherwise we could allow both... and just warn we need that flag if we don't want to overwrite it and still install everything.

It would be good enough for me.

stefk:
> It could be done during composer update but it wouldn't allow to perform a clean install after an update (unless you change it manually).
While it's not a problem for production, it's very annoying in dev.

I don' t understand.

ngodfraind:
If you need to often install your platform and disabled some plugins a little time ago and want/need a proper install, you need to edit your bundles.ini.

Also, from my point of view (tech support guy), if something goes wrong, I'd rather tell to append '--env=maintenance' at the end of a command line than copy that file, editing it, using the command, restoring it if it worked.

Errors do happen and stuff like that is annoying. It won't be if no plugin is disabled tho.

stefk:
Another solution maybe: find a way to hook into the bundle manager which collects/instantiates bundles from bundles.ini, and make it ignore the true/false flags. If this can be acheived programmatically during the update/install, it would way better than using unreliable env magic.

stefk:
You already do that hook. Why not reusing the `.update` flag instead of the environment to detect maintenance mode?

ngodfraind:
Not sure. This would work really well after a composer:update (much easier than copy/paste actually) but the .update flag isn't always here.
Or, we could change the installation process.
Instead of going "claroline:install", we could do 'composer install-claroline'. This would add the .update file and everything would be fine.
I like this tbh.

ngodfraind:
Yes, let's proceed that way:
The "official" way to install the plateform would be through the composer script.
composer install-claroline
composer update-claroline

They would add the .update flag and it would replace the environement. That way, there would be no problem 'ever'. But it would still require an other change in the root composer of the project (wich I don't like either lol).

stefk:
> That way, there would be no problem 'ever'.

Quote of the day.

ngodfraind:
Indeed. The quotes do everything.
But what do you think about that ?

stefk:
Not sure what's the real benefit. If the claroline:install/update commands start by creating the .update flag the result is the same.

ngodfraind:
I'll give it a try. But I thought the kernel was already compiled with the wrong plugin list at that point.

ngodfraind:
It works.
The procedure would be:
```
rm -rf app/cache/*
php app/console claroline:update
rm -rf app/cache/*
```
It's only one more operation that before.

We need to remove the cache the first time to force the initialization of every bundle.
We need to remove the cache the 2nd time to force the initialization of what we really want.

There is still one unresolved issue. In the PluginManager, we want to fetch data from every plugin.
Some of them are stored in their respective bundle class (ie, the requirements). In order to fetch these elements, I also switch the kernel to the maintenance mode. However, it does work here because there is no cache issue but it's not pretty.

Any idea ?

stefk:
I hadn't thought about getting data from disabled bundles, that's a big issue... Can't we just use the bundle manager programmatically when we need it (a slightly modified version of `getActiveBundles` could return all the bundle instances)?

main/kernel/Manager/BundleManager.php
```diff
@@ -48,10 +48,12 @@ public function getActiveBundles()
         $configProviderBundles = array();
         $nonAutoConfigurableBundles = array();
         $environment = $this->getEnvironment();
+        $updateMode = file_exists($this->kernel->getRootDir() . '/config/.update');
+        var_dump($updateMode);
```
stefk: !

ngodfraind:
I think it would work with some tweaking

ngodfraind:
It starts to look good.
The only missing parts are in the front end interface afaik.
I think I will be able to end it tomorrow.

ngodfraind:
@stefk 

Why is the coding check failing ?

stefk:
The diff is in the build log ([here](https://travis-ci.org/claroline/Distribution/builds/123849498#L2368)) but the errors don't seem to be coming from your changes...  Your branch is out-to-date with the master, I'm going to sync it to see if that's the problem.

ngodfraind:
Other than that, there is only the need for a js linter (wich we don't have yet so I assume we don't care yet either).
All these commits can be squashed into one.

I'm more or less done with it.

stefk:
> Your branch is out-to-date with the master, I'm going to sync it to see if that's the problem.

Nope, apparently the #142 pass didn't catch all the errors (maybe the input was too large IDK...). Anyway, these errors are easy to fix (missing blank lines before namespace statements).

stefk:
> It has some tests 

You should add them (and every working test you know) in the [test suite](https://github.com/claroline/Distribution/blob/master/phpunit.xml#L20).

ngodfraind:
They won't pass on travis because it's done after a claroline:install (it requires the list of plugin in the database).

stefk:
You can update the travis config to make it work (we already decided to make a full install in test env anyway). Replacing lines 50-51 of .travis.yml with the following should be enough:

```yml
- php app/console claroline:install --env=test
```

ngodfraind:
When I run the test suite locally:

SYMFONY_DEPRECATIONS_HELPER=weak bin/phpunit -c vendor/claroline/distribution/phpunit.xml 
PHPUnit 4.8.24 by Sebastian Bergmann and contributors.

PHP Fatal error:  Class 'Mockery' not found in /home/nico/Documents/NewDistro/Claroline/vendor/claroline/distribution/main/migration/Tests/MockeryTestCase.php on line 20

ngodfraind:
Travis failed on assetic:dump here: /home/travis/build/Claroline-monolithic-build/app/../web/vendor/innovapath/wizards.css

stefk:
For mockery, see claroline/Claroline#783. Build error is trivial (path should point to web/packages, not frontend bundle).

ngodfraind:
Sure, but why does if only fail now ? This is weird.

ngodfraind:
Too many tests fail when tests are run from a claroline:install. This is not related to this pr. I'm rolling back this change and I'll fix that in an other one.

ngodfraind:
@stefk 
app/phpunit.xml should be removed to avoid confusion. This old file is now useless afaik.

stefk:
The path error is due to the way the build is done. Instead of being in vendor, sources of the repo are in a separate directory, and composer make a symbolic link to them (it was the simplest way to do it).

stefk:
> app/phpunit.xml should be removed to avoid confusion. 

Agreed.

ngodfraind:
The linter has a weird behaviour. Travis finds errors but the claroline:tool command can't find any.

stefk:
Just checked php-cs-fixer:

- version used in travis build is 1.11.2
- my local bin/php-cs-fixer is 1.11.2
- my global php-cs-fixer is 1.11.2
- all share the exact commit/revision

But their result varies, dependending on which one is used and what is the scope of the fix. It makes no sense...

Anyway, using the global version directly on each incriminated file works for me:

    php-cs-fixer fix vendor/claroline/distribution/plugin/collecticiel/InnovaCollecticielBundle.php --dry-run --diff

ngodfraind:
Could you change the dev bundle in order to reflect that ?

stefk:
I don't see any solution unfortunately. The list of files depends on the git index. The best way to be sure we are in sync with the build behaviour would be to use the global version plus a pre-commit hook. 

That makes claroline:tool and the scripts in dev/Resources/scripts/tools useless (the code base is too big anyway).

ngodfraind:
Travis is now happy with that.

ngodfraind:
We can have a discussion about the plugin management interface later, but it fixes the installation/upgrade/update errors so I think it can be merged now.

LaurentGruber:
LGTM

main/core/ClarolineCoreBundle.php
```diff
@@ -47,7 +47,7 @@ public function build(ContainerBuilder $container)
 
     public function supports($environment)
     {
-        return in_array($environment, array('prod', 'dev', 'test'));
+        return in_array($environment, array('prod', 'dev', 'test', 'maintenance'));
```
stefk: I guess `maintenance` env is not needed anymore.

main/core/Manager/PluginManager.php
```diff
@@ -0,0 +1,348 @@
...
+    }
+
+    public function getPluginByShortName($name)
+    {
+        return $this->pluginRepo->findPluginByShortName($name);
+    }
+
+    public function isLoaded($plugin)
+    {
+        $bundles = parse_ini_file($this->getIniFile());
```
stefk: This method is called in foreach loops, meaning the file will be opened and parsed on each iteration. A cache property is needed IMO.

main/core/Manager/PluginManager.php
```diff
@@ -0,0 +1,348 @@
...
+        $errorCount = 0;
+
+        if (array_key_exists('extension', $errors)) {
+            $errorCount += count($errors['extension']);
+        }
+        if (array_key_exists('plugin', $errors)) {
+            $errorCount += count($errors['plugin']);
+        }
+
+        return $errorCount > 0 ? false : true;
```
stefk: `return $errorCount === 0;`

main/core/Manager/PluginManager.php
ngodfraind: A simple cache won't be enough, but   know what do do.

main/core/Manager/PluginManager.php
```diff
@@ -0,0 +1,348 @@
...
+                false,
+                $this->kernelRootDir.'/config/bundles.ini'
+            );
+
+        return $plugin;
+    }
+
+    public function getEnabled($shortName = false)
+    {
+        $bundles = parse_ini_file($this->iniFile);
```
stefk: Same here, cache needed.

main/core/Repository/AdministrationToolRepository.php
```diff
@@ -12,9 +12,32 @@
...
-class AdministrationToolRepository extends EntityRepository
+class AdministrationToolRepository extends EntityRepository implements ContainerAwareInterface
 {
+    private $bundles = [];
+
+    public function setContainer(ContainerInterface $container = null)
+    {
+        $this->container = $container;
+        $this->bundles = $this->container->get('claroline.manager.plugin_manager')->getEnabled(true);
+    }
```
stefk: I already said what I thought about this (tldr: lazy/bad) but if it's the best we can have, at least the `$container` property should be declared.

main/core/Manager/PluginManager.php
ngodfraind: I remember why I did this. After an update, I return the whole plugin list with their dependencies updated as a json list. Everything needs to be refreshed everytime we call these methods otherwise the result may be wrong.

It's fixable tho.

main/core/Resources/modules/plugins/Controller/PluginController.js
```diff
@@ -0,0 +1,44 @@
+export default class PluginController {
```
stefk: This file use hard tabs.

main/core/Manager/PluginManager.php
stefk: I just meant delegating to an internal method which populates a property, like: 

```php
private $bundles;

private function getBundles() 
{
    if (!$this->bundles) {
        $this->bundles = parse_ini_file($this->getIniFile());
    }

    return $this->bundles;
}
```

main/core/Resources/modules/plugins/Controller/PluginController.js
ngodfraind: And we need a js linter.
Wait for it imo.

ngodfraind:
All fixed excepted js tabs.
I'd rather wait our js linter because it's pretty annoying to do manually and it's going to be run on the whole project anyway.

ngodfraind:
@stefk 
Travis is ok.

stefk:
I agree on the js linter but I'm afraid we won't have a tool like php-cs-fixer. The nearest is https://github.com/maxogden/standard-format but it wasn't very reliable in the tests I did. 

BTW there's 6 more uninitialized `$container`.

Otherwise, LGTM

ngodfraind:
There is a package for atom. I'll try it.
Also I'll look for an atom csfixer package.

ngodfraind:
The atom package did work in my case. So it's now formatted using standard style.

ngodfraind:
LGTM

PR #127, opened 2016-04-11
Import formula plugin

ngodfraind:
Import FormulaPluginBundle


LaurentGruber:
LGTM

stefk:
Shouldn't mathjax be fetched with bower?

ngodfraind:
Probably (https://github.com/claroline/Distribution/blob/import-with-history/plugin/formula/README.md)

ngodfraind:
I installed locally from an update and it worked. This is great. Some bugfixes concerning the install are on the disable-plugin branch tho (but it's nothing more than toying with the bundles.bup.ini file).

ngodfraind:
I'll add it in the root readme file.

stefk:
@ngodfraind, @ptsavdar, it could be a good idea to test it with bower now, if we can avoid adding +1000 superfluous files to the repo history.

ngodfraind:
I don't get how it works.

You can install it with `bower install MathJax`

stefk:
I think I got it. This [listener](https://github.com/iCAPLyon1/FormulaPluginBundle/blob/master/Listener/FormulaPluginListener.php#L30) injects this [script](https://github.com/iCAPLyon1/FormulaPluginBundle/blob/master/Resources/views/Formula/plugin.js.html.twig), which includes this [file](https://github.com/iCAPLyon1/FormulaPluginBundle/blob/master/Resources/public/tinymce/plugins/formula/index.html) in a iframe. So maybe using the bower path for MathJax in the last one could be enough?

ngodfraind:
I could not find the last file in the package/MathJax/... directories. That
was my problem.

2016-04-11 15:07 GMT+02:00 Stéphane Klein <notifications@github.com>:

> I think I got it. This listener
> <https://github.com/iCAPLyon1/FormulaPluginBundle/blob/master/Listener/FormulaPluginListener.php#L30>
> injects this script
> <https://github.com/iCAPLyon1/FormulaPluginBundle/blob/master/Resources/views/Formula/plugin.js.html.twig>,
> which includes this file
> <https://github.com/iCAPLyon1/FormulaPluginBundle/blob/master/Resources/public/tinymce/plugins/formula/index.html>
> in a iframe. So maybe using the bower path for MathJax in the last one
> could be enough?
>
> —
> <https://github.com/claroline/Distribution/pull/127#issuecomment-208331558>
>



-- 
Godfraind Nicolas
5, rue de la Carrière
6592 Monceau-Imbrechies
GSM: 0472 94 80 99

LaurentGruber:
What do we do ?

stefk:
I guess at least we should wait for @ptsavdar's input. But merging this as is would be an error IMO.

stefk:
Needs an entry in composer.json#replace (see #137).

LaurentGruber:
Panos travail dessus

ngodfraind:
LGTM

ptsavdar:
LGTM

ngodfraind:
LGTM

PR #128, opened 2016-04-11
Import pdf generator

ngodfraind:
This could be added in the core section too because it no longer needs to be a plugin and there is not much code here.


import.php
```diff
@@ -20,7 +20,7 @@
 cmd("git checkout {$bundle}");
 cmd("git pull --no-tags {$bundle} master");
 cmd("git filter-branch -f --msg-filter 'sed \"1 s/^/[{$bundle}] /\"' HEAD");
-cmd("git checkout import-with-history");
+cmd("git checkout import-pdf-generator");
```
stefk: Probably not intended.

ngodfraind: Yes but it's ok. I usually overwrite this line to do the imports and have to commit the change otherwise the script will crash.

stefk:
LGTM

ngodfraind:
LGTM

ngodfraind:
LGTM

Is it possible to remove the LGTM check if it has already been checked after a merge from the master ?

stefk:
I don't think so, unfortunately.

PR #131, opened 2016-04-11
Video recorder

pitrackster:
Hello ! 
I would like to propose the video recording to the distribution bundle.

The purpose of this plugin is to allow anyone to record his webcam + microphone stream and create a Claroline File resource from this stream.

This plugin make use of libavconv (instead of ffmpeg) and uses MediaRecorder API which have some limitations

Below some screenshots:

##### Before any get user media acceptance
video_recorder_base

##### Access to user media accepted
video_recorder_gum_ok

##### Recording started

video_recorder_recording

##### Recording ended
video_recorder_recording_end

##### The resource created
video_recorder_resource


LaurentGruber:
@claroline/admins should be merge to a separeted branch to test before merging in master ?

ngodfraind:
It's already a separated branch. 

It can me merged imo. Travis check passed, so it's fine for me.

ngodfraind:
By the way, there is a typo in your screenshot ("Lenregistrement") and some translations are missing.

ngodfraind:
LGTM

ngodfraind:
@stefk 
I noticed you changed the travis check to speed up things. It's great but what can we do in this case ?
The installation isn't fired here wich is really the only point we should check.

stefk:
@ngodfraind, the bundle is installed during the build: https://travis-ci.org/claroline/Distribution/builds/122256714#L2902. The change I've made on travis is just to avoid triggering a build for the pr *and* for the push.

ngodfraind:
I'd like to add a few things:
- How do you make it work properly ? (it crashes at the submission file for me, probably because I don't have the required libavtools).
- It should be disabled if the required tools aren't ready (see claroline/distribution#125)
- A lot of translations are missing

ngodfraind:
I just did the merge.

stefk:
@pitrackster, this looks nice but a phpcs-fixer pass would be greatly appreciated ;) I didn't have the time to integrate it as a required status check.

pitrackster:
@stefk yep you are right I'll do it
@ngodfraind : 
- Yes you need libav-tools to make it work
   - the only way I know to check if this lib is there to enable/disable the plugin is during the container compilation, but container (i.e. services) are not available during compilation so I do not see how to use your "disablePlugin" methods.
  - the only "solution" I found is to rewrite the bundles.ini file and set the proper entry to false, IMO this is not a good solution

- Missing translations : I found only one translation missing in warning_message_record_end.en.yml  where are the other missing translations ?

pitrackster:
@stefk do you want me to pass phpcs-fixer on the whole distribution bundle or only on my code ? (it seems that  a lot of plugins need it)

stefk:
@pitrackster, your code is enough IMO. In the future we'll probably have a CS check on PRs, so the codebase will be cleaned progressively and other devs will learn from that process.

ngodfraind:
I guess I didn't clear my cache properly then.

pitrackster:
@stefk done
@ngodfraind do you think I can check that libav-tools is installed in the postInstall() method of my AdditionalInstaller class and if libav-tools is not here disable plugin via ```BundleManager->disable($plugin)```  method ?

If its ok for you I need [this PR](https://github.com/claroline/Distribution/pull/125) to be merged ;-)

stefk:
@pitrackster, I don't think you've pushed your changes.

ngodfraind:
It could really well work, but I'm not done yet with the disable/enable
plugin thing.
I should do this for the ldap aswell (it requires php5-ldap).

2016-04-12 11:51 GMT+02:00 Guillou Patrick <notifications@github.com>:

> @stefk <https://github.com/stefk> done
> @ngodfraind <https://github.com/ngodfraind> do you think I can check that
> libav-tools is installed in the postInstall() method of my
> AdditionalInstaller class and if libav-tools is not here disable plugin via
> BundleManager->disable($plugin) method ?
>
> If its ok for you I need this PR
> <https://github.com/claroline/Distribution/pull/125> to be merged ;-)
>
> —
> <https://github.com/claroline/Distribution/pull/131#issuecomment-208824019>
>



-- 
Godfraind Nicolas
5, rue de la Carrière
6592 Monceau-Imbrechies
GSM: 0472 94 80 99

stefk:
The problem with the `postInstall` method is that there would be no way for a platform admin to enable the plugin later (e.g. after noticing the required lib was missing).

pitrackster:
@stefk yep I was waiting to @ngodfraind answer (and the new plugin manager methods) to make just one PR update... 
I think there will be an interface to enable / disable any installed plugin... But I might be mistaken !

ngodfraind:
Yes, an interface is provided to enable/disable plugins.

Also, even is a plugin is disabled, installation and updates should still
be done (otherwise side effects will occur after the plugin is enabled
again).

2016-04-12 12:03 GMT+02:00 Guillou Patrick <notifications@github.com>:

> @stefk <https://github.com/stefk> yep I was waiting to @ngodfraind
> <https://github.com/ngodfraind> answer (and the new plugin manager
> methods) to make just one PR update...
> I think there will be an interface to enable / disable any installed
> plugin... But I might be mistaken !
>
> —
> <https://github.com/claroline/Distribution/pull/131#issuecomment-208828085>
>



-- 
Godfraind Nicolas
5, rue de la Carrière
6592 Monceau-Imbrechies
GSM: 0472 94 80 99

pitrackster:
@stefk @ngodfraind should I wait for the other PR to be merged or should I push my existing corrections ?

stefk:
You can push your changes, but if they depend on #125, the PR build will break and the review will be way harder.

pitrackster:
yep so I'll wait... I think it's better !

stefk:
Needs an entry in composer.json#replace as well (see #137).

ngodfraind:
What should I check before activating the bundle ?

Any line of php wich will return true or false depending on the lib needed ?

stefk:
@ngodfraind, is the check performed by the core itself? What about checking particular conditions the core doesn't about?

ngodfraind:
The condition is checked by the core itself.
It will search file a "require.json" file at the plugin root and parse it. It currently support php extension but I can improve it for that use case.
It will need to support plugin requirements aswell (some plugins depends on each others).

stefk:
@ngodfraind, answering in #125.

ngodfraind:
@pitrackster 

Everything looks fine on the master branch.
You can run your checks and fire the "disable" method.

$plugin = $om->getRepository('ClarolineCoreBundle:Plugin')->findOneByName('xxx');
$this->container->get('claroline.manager.plugin_manager')->disable($plugin);

pitrackster:
@ngodfraind 

I'm trying to "check plugin required libs and disable the plugin if needed" doing this : 

```
public function postInstall()
  {
      $om = $this->container->get('claroline.persistence.object_manager');
      // load plugin default data
      $default = new DefaultData();
      $default->load($om);

      // check libav-tools is installed
      $cmd = 'avconv -i';
      exec($cmd, $output, $return);
      // if error
      if(count($output) === 0 || $return !== 0){
        echo 'libav-tools not found... disable InnovaVideoRecorderPlugin';
        $plugin = $om->getRepository('ClarolineCoreBundle:Plugin')->findPluginByShortName('InnovaVideoRecorderBundle');
        $this->container->get('claroline.manager.plugin_manager')->disable($plugin);
      }
  }
```
bundles.ini file look like this after running the plugin install command with the requirements not satisfied (i.e. libav-tools not installed) : 

```
...
Icap\WikiBundle\IcapWikiBundle = 1
Icap\OAuthBundle\IcapOAuthBundle = 1
Claroline\ResultBundle\ClarolineResultBundle = 1
Innova\PathBundle\InnovaPathBundle = 1
Innova\VideoRecorderBundle\InnovaVideoRecorderBundle = 
```

IMO each value should be [true] instead of [1] and a disabled plugin value should be set to [false] 
 
Next I tried to install the plugin with libav-tools installed on my computer.

Since the plugin is already installed I need to uninstall it right ?

If I try to uninstall the plugin I get the following error : 
```
  [InvalidArgumentException]                                                                           
  Bundle "InnovaVideoRecorderBundle" does not exist or it is not enabled. Maybe you forgot to add it   
  in the registerBundles() method of your AppKernel.php file? 
```
So I need to enable the plugin by editing the bundles.ini file like this :
 
```
Innova\VideoRecorderBundle\InnovaVideoRecorderBundle = 1
```

Then run the uninstall command to be able to run the install command again... 
Am I doing things the right way? 
- If yes IMO the whole process is a bit harsh 
- If no can you tell me a better way to do this ?

Thanks !

ngodfraind:
I think you can enable it from the administration plugin page. There is no need to reinstall.

ngodfraind:
I'm not sure the claroline:plugin;uninstall command should stay with the distribution package.

pitrackster:
Here is what I've got from the administrate plugin page... Am I in the right place ? 

plugin_admin

What if the platform admin enable the plugin via this interface ? 
Does this action trigger an event that every plugin that need specific libs should implement ?

ngodfraind:
npm run watch / npm run webpack ?

2016-04-25 14:11 GMT+02:00 Guillou Patrick <notifications@github.com>:

> Here is what I've got from the administrate plugin page... Am I in the
> right place ?
>
> [image: plugin_admin]
> <https://cloud.githubusercontent.com/assets/6024701/14783198/7f06c4d4-0aee-11e6-9902-1d4921a5e109.png>
>
> What if the platform admin enable the plugin via this interface ?
> Does this action trigger an event that every plugin that need specific
> libs should implement ?
>
> —
> <https://github.com/claroline/Distribution/pull/131#issuecomment-214286551>
>



-- 
Godfraind Nicolas
5, rue de la Carrière
6592 Monceau-Imbrechies
GSM: 0472 94 80 99

pitrackster:
@ngodfraind ... damn it ! Yep it works much better ;-)

ngodfraind:
I'm not it.

pitrackster:
great !

ngodfraind:
@pitrackster 

The event is done, but there is no easy way to display the requirements in the gui if the bundle is disabled (the event won't be triggered).

ngodfraind:
@pitrackster 
@stefk 

take a look #197

ngodfraind:
@pitrackster 
#197 is merged

```
    public function getExtraRequirements()
    {
        return array(
            'libav-tools' => array(
                'test' => function () {
                    $cmd = 'avconv -h';
                    exec($cmd, $output, $return);

                    return count($output) > 0 && $return === 0;
                },
                'failure_msg' => 'libavtools_explain_message',
            ),
        );
    }
```

This should do the trick.

LaurentGruber:
@pitrackster it's ok for you ? we can merge ?

pitrackster:
@LaurentGruber for me it's ok ! The PR is only missing approvals.

LaurentGruber:
LGTM

PR #133, opened 2016-04-12
Fixing ws suppression if no workspace root

ngodfraind:
There no point doing a `$root->getName()` before checking if $root actually exists. Also it cause issues if we want do remove a workspace and something went wrong before because the $root doesn't exists anymore.


ngodfraind:
LGTM

stefk:
LGTM

PR #134, opened 2016-04-12
Fixing stats user count.

ngodfraind:
It's better to count the number of users directly than incrementing a counter that count the number of users for each roles... because if a user is in 2 different roles, the counter will be wrong.

### DIFF
--- a/main/core/Repository/UserRepository.php
+++ b/main/core/Repository/UserRepository.php
@@ -424,7 +424,7 @@ class UserRepository extends EntityRepository implements UserProviderInterface
     public function countUsers()
     {
         $qb = $this->createQueryBuilder('user')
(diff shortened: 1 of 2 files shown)


ngodfraind:
LGTM

main/core/Repository/UserRepository.php
```diff
@@ -421,6 +421,15 @@ public function countUsersByRole($role, $restrictionRoleNames)
         return $query->getSingleScalarResult();
     }
 
+    public function countUsers()
+    {
+        $qb = $this->createQueryBuilder('user')
+            ->select('COUNT(DISTINCT user.id)')
```
stefk: DISTINCT shouldn't be needed (primary key).

stefk:
LGTM

PR #136, opened 2016-04-12
Avoid name collision with webpack.

ngodfraind:
Otherwise, in the distribution, nobody can share the same entry keys in the assets.json file.

There is no name collision in the files generated, so it only removes this limitation.


ngodfraind:
LGTM

stefk:
LGTM

PR #137, opened 2016-04-12
Add replace section in composer.json

stefk:
This would avoid clashes if an individual package is part of the dependencies along with this distribution.


ngodfraind:
LGTM

LaurentGruber:
LGTM

stefk:
LGTM

PR #139, opened 2016-04-12
Fixing upgrade

ngodfraind:
Upgrade from core-bundle to distribution-bundle fixed (or at least less buggy).

### DIFF
--- a/main/core/Library/Installation/OperationExecutor.php
+++ b/main/core/Library/Installation/OperationExecutor.php
@@ -38,7 +38,7 @@ class OperationExecutor
      * When doing an upgrade from the old core-bundle packages to the distribution one, it will trigger every updater
      * starting from the VERSION_UPGRADE_DISTRIBUTION constant.
      */
-    const VERSION_UPGRADE_DISTRIBUTION = '6.0.0'
+    const VERSION_UPGRADE_DISTRIBUTION = '6.0.0';
 
     use LoggableTrait;
 
@@ -155,7 +155,7 @@ class OperationExecutor
                         $operations[$bundle] = new Operation(Operation::UPDATE, $currentPackage, $bundle);
                         $previousPackage = $this->findPreviousPackage($currentPackage);
 
-                        if ($previousPackage === null && $currentPackage === 'claroline/distribution') {
+                        if ($previousPackage === null && $currentPackage->getPrettyName() === 'claroline/distribution') {
                             /*
                              * Allows easy upgrade from the 6.x core-bundle branch
                              * Note that it will upgrade every plugin that's included in the distribution aswell
@@ -164,7 +164,7 @@ class OperationExecutor
                              * An upgrade from the 5.x branch should first go until the last 5.x version,
                              * then go to the distribution package (it's a bit annoying but it should work).
                              */
-                            $operation[$bundle]->setFromVersion(self::VERSION_UPGRADE_DISTRIBUTION);
+                            $operations[$bundle]->setFromVersion(self::VERSION_UPGRADE_DISTRIBUTION);
                         } else {
                             $operations[$bundle]->setFromVersion($previousPackage->getVersion());
                         }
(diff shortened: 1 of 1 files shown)


ngodfraind:
LGTM (with commit squashed probably)

stefk:
Not sure I understand this, especially the fixed version and the difference between 5.x and 6.x.

With previous-installed.json, you know exactly what bundles were installed and at which version. Isn't it enough to build a precise list of update operations?

ngodfraind:
I guess I can give you a more detailled explanation here. 

I changed the installation/update. It does a backup of the bundles.ini file in order to know what's already installed or not.

As you said, we can compare the distribution-bundle "bundles" section but it obviously doesn't exist for the 6.x branch. The bundle.ini file however is still here and the update is based on that in order to make an easy upgrade.

This way, I can only change the main composer.json file and use claroline/distribution instead of everything, run composer update and everything will work fine.

However, it's a bit hard to know what were every bundle version in that case (altough it's still stored in the previous-installed.json) so I fire every updater from the 6.x branch for them during the first upgrade at the moment. It's a bit arbitrary but it should allow anyone to keep their current installation and upgrade to the distribution without any trouble.

Also (but it's in the plugin enable/disable branch) i'm trying to remove the installFromKernel method because we have 2 differents path for the installation and maintenance would be easier if it wasn't the case. In the master version, claroline:install is an alias of claroline:update (wich looks for the bundles.bup.ini file to performs required operations). It's wrong if you're doing a reinstall because it should be empty to install bundles but it won't be the case and you'll have to remove it. This is already done in the plugin enable/disable branch. 

So we can do it through parsing of the installed.json file but it requires a lot more work and I wanted to make it work asap (so everyone can start to work).

Long story short: it's easier a lot easier to composer our bundles.ini file than composer autogenerated installed.json to find differences and compute update operations. We can switch later to the old system.

ngodfraind:
Also please note that all I said about the bundles.ini usage is only valid for meta packages. Otherwise, it's still the old method and nothing changed.

ngodfraind:
The algorithm currently goes like this:
1) look for the installed.json file
2) find the "bundles" key if it exists
3.a) the bundles key exists: we look for every bundles in that key and we compare them with what's in the backup bundle.ini file => we can compute our update (we could compare with the prev-installed.json aswell but it's harder)
3.b) there is no bundle key: this is not a meta package and proceed as usual

ngodfraind:
The 6.0.0 arbitrary stuff is because with the meta package comparison (with bundles.ini) we can't know the version. Also everyone is going to share the same number version aswell, so I fire everything and that's ok.

stefk:
Ok, thanks for the explanation, that's roughly what I understood. My point concerns 3.a: if you use previous-installed.json instead of bundles.bup.ini, you have all the all the information you need to make the whole thing work both with 5.x and 6.x, without arbitrary decisions. How much harder is it?

ngodfraind:
I'd say we have to rewrite the composer.json "bundles" keys to have something that we can match to the old composer.json because there is nothing allowing us to do that now.

Also this is a bit weird to store those old informations here...

"bundles": [
    "Claroline\\CoreBundle\\ClarolineCoreBundle": "claroline/core-bundle",    
    "My\\New\\Bundle": ""
    .... 
] 

This looks weird to me.

ngodfraind:
By the way, i'm not going to change this right now (I'd rather do the enable/disable plugin stuff before) so I don't think there is anything wrong allowing this pr to pass.

stefk:
Ok, this is a bundle/package mapping issue. I agree that putting the mapping in `bundles` is lame. Another solution (still a bit weak maybe) would be to make a simple string translation in the executor:

```
Claroline\CoreBundle\ClarolineCoreBundle`
Claroline\CoreBundle
claroline/Core-Bundle
claroline/core-bundle
```
 AFAIK, all the bundles in the distribution allow that. 

Anyway, if it's not an emergency, there's nothing wrong IMO in leaving this open. This method is central to the upgrade process and should probably be tested with different scenarios before landing in the master (not to mention it could use a little refactoring).

ngodfraind:
Fixed in #125

PR #154, opened 2016-04-15
[CoreBundle] - Remove flash fallback from embed templates

pitrackster:
Remove flash fallback from resource embed templates to avoid heratic media autoplay.
Also removed the comment from ```plugin/video-player/Controller/VideoPlayerController.php
```


ngodfraind:
LGTM

stefk:
@ngodfraind, you should wait for build completion before approving. @pitrackster, I think there's a superfluous `use` statement, the build is likely to fail because of that.

stefk:
https://travis-ci.org/claroline/Distribution/builds/123342339#L2202

stefk:
LGTM

PR #159, opened 2016-04-18
Remove cs scripts and tool command

stefk:
Applying php-cs-fixer (or any dev script) on a single bundle doesn't make much sense anymore. It's also really not performant. This PR removes the claroline:tool command and the associated cs tools.

Recommended way of integrating php-cs-fixer should be through a pre-commit hook.


ngodfraind:
The translation fixer tool will require some fixing too (or it should be removed aswell).

stefk:
There's a lot of things we could/should do with this bundle (like moving resources/commands from elsewhere). It might be better to track that in a dedicated issue.

LGTM

ngodfraind:
LGTM

PR #163, opened 2016-04-18
[PortfolioBundle] Suggest empty configs for required innova bundles

stefk:
Without the exo bundle that loads them all (https://github.com/UJM-dev/ExoBundle/blob/master/UJMExoBundle.php#L35), no suggested config is provided for some innova bundles that this bundle depends on. They are thus left out of the kernel by the bundle manager, leading to various bugs.


LaurentGruber:
LGTM

ngodfraind:
LGTM

ngodfraind:
LGTM

PR #171, opened 2016-04-19
Optimize test suite memory usage

stefk:
Following [this article](http://kriswallsmith.net/post/18029585104/faster-phpunit), I can see a huge performance gain when running the test suite locally (80% faster, 500% drop in memory).


ngodfraind:
LGTM

stefk:
Before: Time: 2 minutes, Memory: 974.75Mb

After: Time: 1.55 minutes, Memory: 213.00Mb

LGTM

Issue #174, opened 2016-04-19
Import WS and Create a WS with model


david42:
- I  imported a WS with a zip file generated by an export, the announcements resourses are not created in the import
- Also, I have a problem to create a WS with a model (the same WS in the first point), I have these logs:
_
[Tue Apr 19 16:17:16.731620 2016] [:error] [pid 7875] [client
161.3.1.101:51358] PHP Notice: Undefined variable: resourceNode in
/webapps/claroline-connect/vendor/claroline/core-bundle/Manager/WorkspaceModelManager.php
on line 807, referer:
https://claroline-connect.univ-st-etienne.fr/web/app.php/workspaces/new/form
[Tue Apr 19 16:17:16.731837 2016] [:error] [pid 7875] [client
161.3.1.101:51358] PHP Fatal error: Call to a member function getName()
on null in
/webapps/claroline-connect/vendor/claroline/core-bundle/Manager/WorkspaceModelManager.php
on line 807, referer:
https://claroline-connect.univ-st-etienne.fr/web/app.php/workspaces/new/form_

I attach the zip file of the WS:
[MODELE_MEDECINE_9_COURS.zip](https://github.com/claroline/Distribution/files/226686/MODELE_MEDECINE_9_COURS.zip)

stefk:
Import/export of announcements doesn't seem to be implemented yet. Maybe we should open a ticket for that feature?

As for the workspace creation, I can confirm the bug is still here but on line 793. It seems that this code has never been tested.

ngodfraind:
+1

I'm on it.

ngodfraind:
Fixed by #178.

PR #178, opened 2016-04-20
[CoreBundle] Fixing model directory copy.

ngodfraind:
The copy is already handled by the resource manager copy method (wich handles recursion).
This code only duplicate every resource when it runs properly.


ngodfraind:
It should fix #174

ngodfraind:
@stefk 

Can you explain the Scrutinizer thing ?

stefk:
I've just made a test build with scrutinizer. It's disabled now.

ngodfraind:
LGTM

stefk:
LGTM

PR #181, opened 2016-04-20
Add phpmd to travis

stefk:
This PR adds a minimal phpmd check -- only unused private methods for now -- to the travis build. The goal is to progressively add rules to:

- eliminate dead code
- eliminate problematic functions like `exit`, `eval`, `var_dump`, etc.
- limit code complexity (class/method length, dependencies)

cc @claroline/commiters

### DIFF
--- a/plugin/collecticiel/Controller/DropController.php
+++ b/plugin/collecticiel/Controller/DropController.php
@@ -4,6 +4,7 @@
  * Date: 22/08/13
  * Time: 09:30.
  */
+
 namespace Innova\CollecticielBundle\Controller;
 
 use Innova\CollecticielBundle\Entity\Correction;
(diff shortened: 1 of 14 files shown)


LaurentGruber:
LGTM

stefk:
I've removed unused private methods from the code base as a starting point. This needs a careful review. @kitan1982, could you check that the scorm listener is ok?

plugin/collecticiel/Controller/DropController.php
```diff
@@ -4,6 +4,7 @@
  * Date: 22/08/13
  * Time: 09:30.
  */
+
 namespace Innova\CollecticielBundle\Controller;
```
LaurentGruber: @claroline/innova could you also check that there isn't problem with the collecticiel ?

kitan1982:
LGTM

I had copied these private methods to a library class and forgotten to delete the original ones.

ngodfraind:
This is ok for the CoreBundle.
LGTM.

LaurentGruber:
LGTM

PR #191, opened 2016-04-25
[CoreBundle] Ini file mananager update

ngodfraind:
Set "true"/"false" instead of "1"/"")


ngodfraind:
related to #131

stefk:
LGTM

ngodfraind:
LGTM

PR #195, opened 2016-04-25
Support versioning of webpack assets (cache busting)

stefk:
Currently, modules generated by webpack have fixed names, forcing users to do a browser refresh each time an update happened. Moreover, for some reason, a simple F5 is sometimes not enough to get a fresh copy of the file. This is a source of confusion.

This PR makes webpack append a version hash to generated asset files, and adds [this plugin](https://github.com/kossnocorp/assets-webpack-plugin) to get the information needed to include those files in templates. The `hotAsset` twig function now returns the asset URL with the version hash, producing a cache busting on client side after each update.


ngodfraind:
If it works it's good to merge imo.
Will it work for prod aswell ?

stefk:
I've tested in both environments, I didn't encounter any problem.

ngodfraind:
LGTM

stefk:
LGTM...

Issue #158, opened 2016-04-18
Remove vendor deps from portfolio


stefk:
The portfolio plugin has a *Resources/js/vendor* with a lot of copy/pasted dependencies. Most of them are already available in *web/packages*.

cc @claroline/icap-ec.

ngodfraind:
Also, the @InnovaXxxBundle aren't found anymore so you get a blank page on the index. I don't know why.

stefk:
@ngodfraind, that bug should be fixed by #163.

stefk:
Reopening, the original problem is still there.

ptsavdar:
Fixed in #196

PR #196, opened 2016-04-25
Removed unnecessary vendors, changed path from FrontEnd to packages

ptsavdar:
Fixing #158


plugin/portfolio/Resources/views/Portfolio/index.html.twig
```diff
@@ -96,7 +96,7 @@
       '@IcapPortfolioBundle/Resources/views/less/comments.less'
       '@IcapPortfolioBundle/Resources/views/less/statistics.less'
       '@IcapPortfolioBundle/Resources/views/less/daterangepicker.less'
-      '@FrontEndBundle/Resources/public/toastr/toastr.less'
+      'packages/toastr/toastr.less'
```
LaurentGruber: is it not better to compile less file in the installation process (or in the package creation process) and reference the compiled file in the template ?
@claroline/maintainers

ptsavdar: The files are compiled during installation/update process then minified and stored in the web folder (as css files). The app then loads these files every time it is called.

LaurentGruber: Ok thanks. LGTM

stefk:
LGTM

LaurentGruber:
LGTM

PR #197, opened 2016-04-25
Plugin event ready checker

ngodfraind:
Allow plugin to perform extra dependencies check in anonymous functions.

How to use
---------------

Go into your bundle class and add the 'getExtraRequirements()' method.
Here is an example.

This is an array of operations to do.
- 'test' is an anonymous function wich will return if the requirements are met
- 'failure_msg' is the message shown when these requirements are missing. It's translated from the 'error' domain.

```
    public function getExtraRequirements()
    {
        return array(
            'libav-tools' => array(
                'test' => function () {
                    $cmd = 'avconv -h';
                    exec($cmd, $output, $return);

                    return count($output) > 0 && $return === 0;
                },
                'failure_msg' => 'libavtools_explain_message',
            ),
        );
    }
```

We have to do something like this because the bundles aren't loaded if they're disabled and we can't listen to events or do anything like this.

These commits can be squashed.


stefk:
That's a good solution, though the naming is a bit weird. Also it would make more sense to have the function return true when the requirement is met IMO.

ngodfraind:
What do you suggest for the naming ?
The function should return true when the requirements are met. It's already the case. I copy pasted the code I got from the other pr and didn't put any more thought on it.

At first, I used a simple tab but firing $requirements[0](); was even worse.

stefk:
> What do you suggest for the naming ?

`getExtraRequirements` maybe? You already have a `checkExtraRequirements` method, that would be consistent. For the keys, it could be `test`/ `failure_msg` instead of `function` / `warning`.

> The function should return true when the requirements are met. It's already the case. 

It doesn't seem so. Both your example and the function [call](https://github.com/claroline/Distribution/pull/197/files#diff-e7ef1e17c440dceed06cdd1035ceb647R378) take a truthy value as an errror.

pitrackster:
@ngodfraind this look good to me !

 ```return count($output) === 0 || $return !== 0;``` 
will be false if libav-tools is installed.

it should be ```return count($output) > 0 && $return === 0;```

ngodfraind:
Every method updated to be consistent.
I tried with the other return and it works (install/uninstall libav-tools)
It looks good to me.

stefk:
LGTM

ngodfraind:
LGTM

PR #202, opened 2016-04-26
[CoreBundle] Fixing some dbwriter typos.

ngodfraind:


stefk:
I can't see any use of these plugin config options. Are you sure there won't be breakage outside this repo?

Otherwise LGTM.

ngodfraind:
It's not used elsewhere yet atm but it was requested.
The cursus bundle should use them sooner or later.
LGTM

PR #226, opened 2016-04-28
Facets update

ngodfraind:
@stefk 
@Elorfin 

The facet refactoring is not done yet. I'm opening this pr to discuss an angular form directive I wrote while doing it.
Making form with binding and validation is pretty annoying, so I wanted to make my life easier but it's still only a draft.

You can look at it [here](https://github.com/claroline/Distribution/tree/facets-update/main/core/Resources/modules/form). A readme is provided.

I'd like to have some input about it, because we'll either make all our form by hand (wich is a possibility) or eventually do something similar. This is something we can already decide imo.


Elorfin:
I will have no time to check this before the weekend, but I will take a look next week.

stefk:
Quick feedback (no time to review it all now):

As the goal is to simplify form building, the API could be made more concise, à la Symfony:
```js
// simplest way to declare a field, just the name and the type:

['foo', 'text']

// with basic config:

['bar', 'checkbox', { 'label': 'Bar' }]

// with more config:

['baz', 'select', {
  validators: ['not-blank'],
  translation_domain: 'platform',
  options: [
    { value: 1, label: 'foo'},
    { value: 2, label: 'bar'},
    { value: 3, label: 'baz'}
  ]
}]  
```

TLDR: `Field = (name, type, options = {})`

Also, ideally, provided field types should have a default config, overridable both at the field and at a the form level (i.e. there're things you probably don't want to repeat again and again in each field).

ngodfraind:
I'll work on the forms tomorrow. The only issue I have (for the future) is 
 validators: ['not-blank'],

I wanted to be able to send some parameters to the validator (not sure how to get them yet).
example: let's assume I need to check that a value is between 1 and 10.
I could make a "rangevalidator" and do something like this:

```
validators: {
  'integer-range': {
    'min': 1,
    'max': 10
  }
}
```

Otherwise I agree

Elorfin:
I have given a look at what you've done for the Forms. Looks cool.

People will maybe want to use the fields outside of the context `<formbuilder>` so we need to think to keep the two perflectly separated.

Another thing is that if we go for a Form library, we will quickly want to add custom field types like `<form-select-resource>` or `<form-select-user>`. 

It seems you bind your validators via HTML5 validation attributes (in the Text field). It will work for basic fields like text or number, but for custom field types it would be too limited.

Why not use the notation `new *` : 

```
validators: {
  new NotEmpty(),
  new NumberRange({ 'min': 1, 'max': 10 })
  'integer-range': {
    'min': 1,
    'max': 10
  }
}
```

**Pros :**
- can handle custom options directly in the object class
- can provide a `validate(Object: data)` method usable by events (ex. `data-ng-change`) or at submitting time by the form.
- can provide a custom message string to populate `$error`. So the `<help-block>` can be called directly into [field.html](https://github.com/claroline/Distribution/blob/facets-update/main/core/Resources/modules/form/field.html) (no need to call it in each type) and has just to loop over the array of validation messages to display it.
- we can also provide directives on top of this classes (eg `data-integer-range="{}"`) if needed.

**Cons :** 
- The `new` synthax is heavier
- We will need to `import` a library in the Controller that define the Form validators.

A way to solve this and keep the raw object synthax :

```
validators: {
  'integer-range': {
    'min': 1,
    'max': 10
  }
}
```
is to create a `ValidatorProvider` which will achieve the mapping between the validator string name (``"integer-range"`) and the validator class (or the validator string may match the validator class name).

Some questions/remarks about the current implementation:
- the label can directly be put in field.html (not in every type)
- What's the purpose of `formName` property in the fields directives ?
- Why the whole FormController (which has defined the form) is passed to the field ? I have seen only one use to get the `$error` array in the HelpBlock. It can be accessed directly from `ngModel` which is available in the field directive. So there is no need to pass the controller to : Field / HelpBlock / each FieldType / Form.

Also it's not a very good practice to pass the whole controller into all directives. If you need some methods from the controller (like onSubmit) directly pass the methods as directive parameters. It will be cleaner and easier to test (if we do frontend tests some day).
 
- Neither need to the field name in the HelpBlock (as you don't need to retrieve the `$error` object from the controller)
- Keep the `validators` into the field object to avoid the extra param `validator` in field directive.

We can also provide a directive to encapsulate the `<form>` tag into a directive (with the onSubmit callback as parameter) for simplier use and more control over the rendering of buttons.

**NB :** 
I'm not sure this is something we can fix in Angular 1 (maybe in Angular 2), but I want to draw your attention on the file [field.html](https://github.com/claroline/Distribution/blob/facets-update/main/core/Resources/modules/form/field.html). It uses a `ng-switch` to choose which field type to display. I don't know if you know it but in this case the Angular process is : 
- load all directives templates through HTTP (and as the template is defined by the directive, both the Directive and Controller are instanciated too)
- append them to the DOM
- remove (but keep track on in the case `ng-switch` value change) the ones that doesn't match the `ng-switch`

This can reduce performance specifically for big forms or/and when we will have many types defined. That can be a problem for mobile user too.

This is the same process which is used in the ExoBundle to choose the QuestionType to display and I think, later, in the PathBundle to choose the Resource to display.

That's a recurring problem and we need to keep an eye on it to not bloat the UI. I have to check if there are proper way to achieve this in Angular 2. I have found nothing for Angular 1 for now :(.

ngodfraind:
API simplified.

ngodfraind:
I'm working on it.

@Elorfin 

- the labels: They're in the type directive because it allows me to use the directive directly without the needing the <formbuilder> one

- formName: good catch. I forgot to remove it

- the formController: I didn't know that. I'll give it a try

ngodfraind:
@stefk 
@Elorfin 

Your suggested changes have been done.

stefk:
@ngodfraind, it looks goods but the PR doesn't seem ready for a review (lot of stuff missing/incomplete).

ngodfraind:
Yes, I've still a lot of work to do. I only wanted to have the good form component for now.

ngodfraind:
@stefk 
@Elorfin 

What do you think about making the form module a separate bower package ? This is pretty generic (and I could already use it to fix some other issues).

It's only missing tests.

stefk:
Why not, but we just merge separate repos (including generic ones) into this one to ease development, so this is a bit of a step back.

Elorfin:
Another point is we will want to add custom types specific to Claroline (select-user, select-resource, selecte-role, etc.). So it will become only usable here.

ngodfraind:
Good point.

2016-05-04 11:21 GMT+02:00 Axel Penin <notifications@github.com>:

> Another point is we will want to add custom types specific to Claroline
> (select-user, select-resource, selecte-role, etc.). So it will become only
> usable here.
>
> —
> <https://github.com/claroline/Distribution/pull/226#issuecomment-216808493>
>



-- 
Godfraind Nicolas
5, rue de la Carrière
6592 Monceau-Imbrechies
GSM: 0472 94 80 99

ngodfraind:
The backend is more or less done.
It's only missing some drag & drop feature.
The front end still need html/css fixes.

ngodfraind:
@claroline/maintainers 

There are still some minor issues (gui not pretty (altough usable), drag & drop not implented everywhere and some translations are still missing afaik).
But overall, it's good enough to be reviewed since there are so many changes and the deadline is so close.

It refactors the facet management to use angular, change the user profile (it merge private & public profile into a single page with a "edit mode") and adds a few field types for the facet.

main/core/Resources/modules/profile/Partial/profile.html
```diff
@@ -0,0 +1,77 @@
...
+      <li ng-if="!facet.is_main" ng-repeat="(key, facet) in pc.facets"> <a href="#facet-{{ facet.id}}" data-toggle="tab"> {{ facet.name }}</a> </i>
+  </ul>
+  <div id="facet-pane" class="tab-content">
+      <div class="tab-pane active" id="general-profile">
+          <div class="panel-body">
+              <div class="row">
+                  <aside class="col-md-2 profile_aside" role="complementary">
+                      <div class="avatar">
+                          <img ng-if="pc.user.picture"
+                               src="{{ 'uploads/pictures/' +  pc.user.picture|asset }}"
```
ngodfraind: Make an angular "asset filter"

main/core/Resources/modules/services/Service/ClarolineAPIService.js
```diff
@@ -79,9 +81,6 @@ export default class ClarolineAPIService {
     }
 
     confirm(urlObject, callback, title, content) {
-        //the order is important
```
ngodfraind: Remove comments

main/core/Repository/ResourceNodeRepository.php
```diff
@@ -178,8 +178,8 @@ public function findChildren(ResourceNode $parent, array $roles, $user, $withLas
         if ($withLastOpenDate && $user !== 'anon.') {
             $this->builder->selectAsArray(false, true)
                 ->whereParentIs($parent)
-                ->addLastOpenDate($user)
-                ->groupById();
+                ->addLastOpenDate($user);
+                //->groupById();
```
ngodfraind: put back group by id

stefk:
@ngodfraind, make the build pass, so that we can deploy and test the gui.

ngodfraind:
@stefk done

main/core/Controller/API/User/FacetController.php
```diff
@@ -0,0 +1,310 @@
...
+     */
+    public function putPanelRoleAction(PanelFacet $panel)
+    {
+        $params = $this->request->request->all();
+        $this->om->startFlushSuite();
+
+        foreach ($params['roles'] as $param) {
+            $role = $this->om->getRepository('ClarolineCoreBundle:Role')->find($param['role']['id']);
+            $canOpen = $param['can_open'] === 'true' ? true : false;
+            $canEdit = $param['can_edit'] === 'true' ? true : false;
```
stefk: Minor, but:
```php
$param['can_open'] === 'true' ? true : false
```
is equivalent to:
```php
$param['can_open'] === 'true'
```
(the `===` operator already returns a boolean value)

main/core/Controller/API/User/FacetController.php
```diff
@@ -0,0 +1,310 @@
...
+    public function putProfilePreferencesAction()
+    {
+        $params = $this->request->request->all();
+        $this->om->startFlushSuite();
+
+        foreach ($params['preferences'] as $param) {
+            $role = $this->om->getRepository('ClarolineCoreBundle:Role')->find($param['role']['id']);
+            $baseData = $param['base_data'] === 'true' ? true : false;
+            $mail = $param['mail'] === 'true' ? true : false;
+            $phone = $param['phone'] === 'true' ? true : false;
```
stefk: Same

main/core/Controller/API/User/ProfileController.php
```diff
@@ -0,0 +1,119 @@
...
+
+        return $facets;
+    }
+
+    /**
+     * @Get("/profile/{user}/links", name="get_profile_links", options={ "method_prefix" = false })
+     */
+    public function getProfileLinksAction(User $user)
+    {
+        //add check access
```
stefk: Still relevant? Security check missing?

ngodfraind: I don't think so. I think it was already public before.

main/core/Entity/Facet/Facet.php
```diff
@@ -133,23 +154,35 @@ public function setRoles(array $roles)
...
+    public function setForceCreationForm($boolean)
     {
-        $this->isVisibleByOwner = $boolean;
+        if ($boolean === 'true') {
+            $boolean = true;
+        } elseif ($boolean === 'false') {
+            $boolean = false;
+        }
+
+        $this->forceCreationForm = $boolean;
```
stefk: These 7 lines can be reduced to 1:

```php
$this->forceCreationForm = $boolean === 'true';
```

(although it's ugly to make that kind of conversion here)

main/core/Entity/Facet/Facet.php
```diff
@@ -133,23 +154,35 @@ public function setRoles(array $roles)
...
+    public function setIsMain($boolean)
     {
-        $this->forceCreationForm = $boolean;
+        if ($boolean === 'true') {
+            $boolean = true;
+        } elseif ($boolean === 'false') {
+            $boolean = false;
+        }
+
+        $this->isMain = $boolean;
```
stefk: Same

main/core/Manager/FacetManager.php
```diff
@@ -511,24 +449,29 @@ public function getFieldFacet($id)
...
+
     public function getFieldFacets()
     {
         return $this->om->getRepository('ClarolineCoreBundle:Facet\FieldFacet')->findAll();
     }
 
-    public function getVisibleFieldFacets()
+    /**
+     * Unused.
+     */
```
stefk: ?

main/core/Repository/PanelFacetRoleRepository.php
```diff
@@ -0,0 +1,18 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Claroline\CoreBundle\Repository;
+
+use Doctrine\ORM\EntityRepository;
+
+class PanelFacetRoleRepository extends EntityRepository
+{
+}
```
stefk: To be removed.

main/core/Repository/ResourceNodeRepository.php
stefk: as a comment? what do you mean?

main/core/Resources/modules/facets/Controller/FacetController.js
```diff
@@ -0,0 +1,461 @@
...
+
+        const qs = this.FormBuilderService.generateQueryString(list, 'ids')
+        this.$http.put(Routing.generate('api_put_fields_order', {panel: panelId}) + '?' + qs).then(
+          d => {
+          },
+          d => ClarolineAPIService.errorModal()
+
+        )
+      }
+    })
```
stefk: 162 lines constructor.

ngodfraind: To be fair, there are around 60 lines of form definitions and 60 more of dragula configuration. Dragula needs to be here (it needs to use $scope events) and form definitions can't be compressed much more.

stefk: That's not a blocking issue, I just mentioned it because ideally, the design should be better. For example, one nice thing about the form component you've added is that form definitions become almost completely declarative, so you can move them to separate files and require them when needed. As for the scope binding, you don't need an extra class, you could simply move the logic in helper methods:
```js
$scope.$on('panel-bag.drop', this.onPanelBagDrop.bind(this))
```

stefk: Of course there might be more urgent things to do now...

ngodfraind:
@claroline/maintainers 
Ready to be deployed and tested again (last test was successfull).

LaurentGruber:
LGTM

ngodfraind:
@claroline/maintainers It looks ok now.

stefk:
LGTM

PR #227, opened 2016-04-28
[CoreBundle] Fixing wrong admin tool check

ngodfraind:
The home_tabs tool doesn't exist so the check will fail for everyone who don't have the ROLE_ADMIN.

desktop_and_home exists and is probably what was intended.


stefk:
LGTM

ngodfraind:
LGTM

PR #245, opened 2016-05-02
[SensioInsight] Fix critical issues

stefk:
None of them actually presented a security risk, but fixing them led to a cleaner, shorter and sometimes more efficient code.

This should get us the bronze medal.


ngodfraind:
This looks good. I'm not comfortable with the exp() function of the query builder so I can't tell if this is correct or not. Have you done a quick test (ie disabling a resource and check if doesn't appear anymore) ?

Otherwise LGTM

stefk:
Yes, I've tested the effect of plugin (de)activation on widgets. I didn't notice any problem.

(LGTM)

PR #248, opened 2016-05-02
[CoreBundle] ini parse_ini_file handling.

ngodfraind:
Set true/false when disabling/enabling plugin in bundles.ini


stefk:
LGTM

ngodfraind:
LGTM

PR #259, opened 2016-05-06
Fix plugin requirements handling

stefk:
Refactors the plugin manager to avoid access to undefined array entries when checking plugin requirements. Also adds a bit of doc and cleans/renames where appropriate.


stefk:
LGTM

ngodfraind:
LGTM

PR #267, opened 2016-05-11
[BundleRecorder] Exclude non symfony bundle classes

stefk:
Removes a false positive in bundle detection, which makes all the travis builds fail:

https://travis-ci.org/claroline/Distribution/builds/129190687#L2036


LaurentGruber:
LGTM

stefk:
LGTM

PR #268, opened 2016-05-11
Add a dependency cache layer in travis builds

stefk:
This should reduce significantly the build time.

@LaurentGruber, this PR is dead without a server access.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,23 @@ language: php
 php:
   - "5.5"
 
+addons:
+  apt:
+    packages:
+      - sshpass
+
 env:
-  - DB=mysql
+  global:
+    # REMOTE_HOST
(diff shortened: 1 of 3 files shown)


stefk:
It works. I've added a step to deploy PRs "previews" to the remote server (i.e. archives installable with the `composer fast-install` command):
- [cache](http://149.202.136.210/cache)
- [previews](http://149.202.136.210/preview)

What we probably need now:
- some cache invalidation/refresh on the remote server (we may start with a simple cron)
-  a simple way to install previews from the PRs (hubot/docker?)

@claroline/maintainers, good to merge IMO.

ngodfraind:
LGTM then

LaurentGruber:
LGTM

PR #270, opened 2016-05-11
[CoreBundle] UserManager:bindUserToOrganization() method performance …

ngodfraind:
…significantly increased for a large amount of users (it's probably over 100x times faster when the number is over 10k).


stefk:
LGTM

ngodfraind:
LGTM

PR #271, opened 2016-05-11
[MessageBundle] Fixing the last message widget behaviour.

ngodfraind:
This widget was broken.
- it displayed messages from every forums as long as you had a workspace access (even if the forum was private)
- the widget configuration didn't work properly (it only show messages from forums you created wich is not what it was supposed to do).

This pr removes the widget configuration (it's tricky to implement with the current database) and fix the 1st issue.

fiy, the configuration was to be able to only show forums you put a comment. It's doable but it looks like I'd have to do a big dirty slow request to achieve that with the current implementation (not worth it imo).


ptsavdar:
Not checked in details but overall LGTM

stefk:
LGTM

PR #277, opened 2016-05-12
[path-bundle] Added tinymce to path description

ptsavdar:


Elorfin:
LGTM

Maybe we can put the `TinymceConfigService` and `trustAsHtml` filter in the CoreBundle.

- The config service will permit to always have the same config for all TinyMCE fields.
- We need to regularly redefine trustAsHtml (we have the same filter in ExoBundle).

ptsavdar:
@Elorfin You're right, me too I've thought that but then I considered that there are bundles and scripts in es6, compiled using webpack and others in es5 so I thought that maybe it would be better if we waited for a more homogenous version. For example I have these functions in WebsiteBundle but in es6. Which means it needs compiling, while in path bundle which doesn't use webpack we need the es5 version of it. Maybe I've thought it wrong, I don't know. What do you think? (But I totally agree with you @Elorfin in a certain point it would be nice if we grouped all these utilities in core bundle)

stefk:
Shouldn't we merge this as is and open an issue to keep track of things we should move to the core (and discuss es5-6 issues)?

ptsavdar:
Ok I'm merging it. LGTM

LaurentGruber:
LGTM

stefk:
17.46 min without cache (https://travis-ci.org/claroline/Distribution/builds/129978101)
6.38 with cache (https://travis-ci.org/claroline/Distribution/builds/129982238)
Win!

PR #296, opened 2016-05-17
Improve empty exercise display (for admins & users)

Elorfin:
This PR enhance the display of Exercise that have not already been configured (aka the Exercise has no steps configured) to have the same behavior as the Paths (as requested by @LaurentGruber).

So the overview of an empty Exercise will look like :

For admins :
empty-admin

For users :
empty-user

And the overview of an Exercise that has at least one step : 
For admins :
configured-admin

For users :
configured-user

I think for regular Users we can completely remove the topbar (as they only have access to the player and the results) and move the results button just above the "Answer" button.


stefk:
LGTM

LaurentGruber:
LGTM

PR #299, opened 2016-05-17
Fix default value of Exercise description

Elorfin:
Set an empty string as the default value for Exercise description (because the validation schema does not allow `null` value).


stefk:
LGTM

purplefish32:
LGTM

LaurentGruber:
LGTM

PR #317, opened 2016-05-17
fix display of 0 score

Elorfin:
If there is no score set in the Paper we display a "?", but if the user obtains 0 to his paper, it was miscatch as undefined.


purplefish32:
LGTM

LaurentGruber:
LGTM

PR #323, opened 2016-05-18
Use --prefer-dist option + fix step order

stefk:
See https://github.com/composer/composer/issues/4884#issuecomment-195229989.


stefk:
LGTM

LaurentGruber:
LGTM

PR #325, opened 2016-05-18
add matching css

charlotte42:
C'est surtout pour tester si on a fait la bonne procédure pour les branches et les PR.


Elorfin:
LGTM

stefk:
LGTM

Elorfin:
@claroline/ujm : in fact the css files was not removed but was moved in Resources/public/less in order to integrate them in the main compiled less file.

No need to fix it for now, but just for you information.

charlotte42:
Ok thanks

Issue #149, opened 2016-04-15
[ExoBundle] Reset timer


Elorfin:
The localStorage is never reset, so if you start a new attempt, the calculated duration is wrong.

Add a TimerService in combination with the TimerDirective, and call a reset() method while Exercise is ended (ExercisePlayerCtrl.end())

Cindy42:
yes it's reset in the ExercisePlayerCtrl.js, in the function handleStepNavigation from line 236 to 241 (in branch v6-UJM).
Do you have this problem after moving the timer into a stand alone directive or before ?

Elorfin:
You only reset the var `counter`, but not the vars that contain hours, minutes, seconds. So when you restart the exercise the calculation is made on the old values because this part of code : 

```
exoPlayer.$localStorage.$default({
        counter: 0,
        hours: 0,
        minutes: 0,
        secondes: 0
});
```

only set values in local storage if they are undefined (and they are defined by the previous attempt).

PR #328, opened 2016-05-18
[ExoBundle] Reset timer

marion42:


Elorfin:
OK, this fix the problem if the exercise is ended because the end of the duration is reached.

But this part of code is never executed (and so the timer is not reset) if : 
- The User ends the Exercise before the duration is down.
- The User interrupts the Exercise with the "Interrupt" button.

marion42:
Yes, this problem was notice yesterday. This is cause by the directive. When i write it, i add a call to a reset in the 'end' if of the validateStep function (when the timer was in the controller). Same should be done with the directive but i couldn't find where ...

Cindy.

Elorfin:
The Timer reset is fixed.

I have created a TimerService to be able to interact with the Timer through JS code (e.g. start or stop the Timer).

The directive has been simplified to only display the Timer object (it's no longer required to initialize a new Timer).

I have also changed the way the data are stored in localeStorage : 
- added a wrapping `timers` object to keep things organized if someone else want to use localeStorage
- when you create a new Timer, you need to provide an ID (In our case, I use the Exo ID). The Timer will be stored in `timers[MY_ID]`), so there is no collision if many Timers are defined (but the current implementation does not manage the case of multiple tabs opened on the same Exercise).

@marion42 : FYI, your first commit seemed to not be necessary because as stated in the [ngStorage documentation](https://github.com/gsklee/ngStorage#delete-everything--demo) the line you have replaced is a shortcut (to clear all the localeStorage) for the version you have added.

@claroline/maintainers : can you review this pull-request please ?

Elorfin:
Fixes #149.

ngodfraind:
The coding conventions are these ones for javascript afaik: [https://github.com/feross/standard](url)

Otherwise LGTM

Elorfin:
@ngodfraind : I can see the repo you point out in our `package.json` but is there really someone who use these standards ? 

For example, all the JS I have seen in the platform have 4 spaces.

Also, I have a problem with theses rules : 
- No semicolons : ok for ES6 but for ES5 I have always read it's better to add them.
- 2 spaces indent : the main argument seems to be to decrease the size of uncompiled JS files => we compile them, so why to not use other languages which uses 4 spaces ?
- Space after function name : personal preference for this one, the PHP syntax is nicier (no space between function name and parameters `function myFunction()` instead of `function myFunction ()`).

The others seems legit.

If I remember correctly @stefk has talked about using ESLint but I can't find any config file in Claroline or Distribution...

ngodfraind:
There is no linter currently. These are the conventions @stefk told me to use (altough it's only es6 atm) altough they're not enforced.

Elorfin:
OK I take note of this rules. 
Thanks for the info.

For now I will just ignore them (to let the code style compliant with the ExoBundle code base). 
I will fix it when we will use ES6 in Exo or when we will have an automated linter.

LaurentGruber:
LGTM

Issue #155, opened 2016-04-15
[CoreBundle] Error handling broken when dealing with templates


stefk:
Happens in v5 when using self-registration, if the default workspace template archive is not available:

> Undefined property: Claroline\CoreBundle\Library\Workspace\Configuration::$data in [...]/vendor/claroline/core-bundle/Claroline/CoreBundle/Library/Workspace/Configuration.php on line 173

There's still no `$data` property currently, and error management in this class seems broken.

ngodfraind:
The default.zip archive should always be available. The error handling could be changed but that's the only thing I can think of.

stefk:
That's not the first time this error appears. I know the archive is *supposed* to be there but the fact is sometimes it's not and that problem shouldn't be hidden behind programming errors.

stefk:
Another pain point: when the unzip binary isn't available, the code fails silentiously at `$archive->open($path)`. No error, no ouput, just like a `die`.

ngodfraind:
What do you think about removing the configuration class ?

stefk:
I don't know, it depends on what the new design would be. On the other hand, archive manipulation with proper error handling is clearly not tied to workspace creation, so that part could/should probably be extracted.

PR #332, opened 2016-05-18
[CoreBundle] Refactoring workspace creation.

ngodfraind:
There are still some minor optimizations that could be done but it's much better overall.

- Fix Blog tag import
- Removes the Workspace\Configuration class
- Adds tests to the workspace API (and fix the old weird broken stuff)
- Much cleaner import overall (no more $this->data and $this->workspace in services).
- A lot of dead code removed
- TransfertManager changed to TransferManager
- Fixing rich text parser for cli import

todo:
- Test the cursus [done]
- double check everything is working [done]
- clean the template directory after import [done]
- Fix the tests for travis


ngodfraind:
@stefk these tests run locally, why do they crash on travis ?
Could it be a permission issue ?

Also, I think we could add a "test" directory at the application root level.

ngodfraind:
Related to #155

main/core/Controller/ResourceController.php
```diff
@@ -1125,28 +1123,17 @@ public function importAction(ResourceNode $directory)
...
+            $user = $this->tokenStorage->getToken()->getUser();
+            $this->transferManager->importResources($template, $user, $directory);
 
+            return new JsonResponse(array());
+        } else {
             return array('form' => $form->createView(), 'directory' => $directory);
-        //}
+        }
+
+        return array('form' => $form->createView(), 'directory' => $directory);
```
stefk: Dead code. Best way to fix it: remove the `else` branch above.

main/core/Library/Transfert/RichTextFormatter.php
```diff
@@ -68,6 +72,7 @@ public function __construct(
         $this->transferManager = $transferManager;
         $this->maskManager = $maskManager;
         $this->eventDispatcher = $eventDispatcher;
+        $this->ch = $ch;
```
stefk: Minor, but the name is cryptic.  If you want something short, `$config` is better IMO.

main/core/Manager/WorkspaceManager.php
```diff
@@ -1152,4 +1119,36 @@ public function getLogger()
...
+        if (!$refresh) {
+            return $this->importData;
+        }
+
+        $archive = new \ZipArchive();
+        $fileName = $file->getBaseName();
+        $extractPath = $this->templateDir.$fileName;
+
+        if (true === $code = $archive->open($file->getPathname())) {
+            $res = $archive->extractTo($extractPath);
```
stefk: The extract path is the same than the archive path?

main/core/Manager/WorkspaceManager.php
```diff
@@ -1152,4 +1119,36 @@ public function getLogger()
...
+
+        throw new \Exception("The workspace archive couldn't be opened");
+    }
+
+    public function removeTemplate(File $file)
+    {
+        $fileName = $file->getBaseName();
+        $extractPath = $this->templateDir.$fileName;
+        $fs = new FileSystem();
+        $iterator = new \DirectoryIterator($extractPath);
```
stefk: That iterator is not used.

main/core/Manager/WorkspaceManager.php
stefk: Shouldn't we deal with extraction errors as well to definitely close #155?

ngodfraind: The extract path is different.

stefk:
@ngodfraind, much better indeed. Regarding the tests, they both fail because the workspace cannot be created due to an extraction path error. Two things to consider:
- my comment above about `extractTo`
- the way default templates are installed now (I saw you removed `LoadTemplateData`, how is it supposed to work now?)

ngodfraind:
They're fetched directly from the Resources/templates directory.

main/core/Manager/WorkspaceManager.php
stefk: Looks like `$extractPath === $templateDir.$templateFileName === $templatePath`

main/core/Manager/WorkspaceManager.php
```diff
@@ -1152,4 +1119,38 @@ public function getLogger()
...
+        if (!$refresh) {
+            return $this->importData;
+        }
+
+        $archive = new \ZipArchive();
+        $fileName = $file->getBaseName();
+        $extractPath = $this->templateDir.$fileName;
+
+        if ($archive->open($file->getPathname())) {
+            $res = ;
```
stefk: !

main/core/Manager/WorkspaceManager.php
ngodfraind: There is no $templatePath var

main/core/Manager/WorkspaceManager.php
ngodfraind: File was not saved properly lol.

main/core/Manager/WorkspaceManager.php
stefk: I know, that's pseudo-code...

`$extractPath === $this->templateDir.$file->getBaseName() === $file->getPathname() // = template path`

Actually that's not true for default templates if they're fetched from the resources directory, but isn't it the case for custom templates?

ngodfraind: It's not even the case because they end up en the /tmp folder. They're not stored. The templates directory currently only act as a temporary thing but the extracted default and personal.zip should stay here as an optimization.

stefk: https://travis-ci.org/claroline/Distribution/builds/131388919#L1773:

*default.zip* seems to be used as a directory, which is weird. Is `extractTo` supposed to create the destination dir?

You should empty your file/templates directory and run the test suite again. This is not a travis issue IMO.

ngodfraind: My directory is already empty. I'll remove it completely and use the /tmp folder instead.

stefk: Sorry, I don't understand. If you remove the templates directory, where will the custom templates be stored?

ngodfraind: They won't. You currently can't pick them from that folder anyway. They're
usually used from the cli interface because they tend to be rather big
(100mo - 1Go) as the current system doesn't show the import status yet. And
also, uploading such a large file is usually a pain. It currently work as a
one shot thing.

2016-05-19 14:16 GMT+02:00 Stéphane Klein <notifications@github.com>:

> In main/core/Manager/WorkspaceManager.php
> <https://github.com/claroline/Distribution/pull/332#discussion_r63867313>:
>
> > @@ -1152,4 +1119,36 @@ public function getLogger()
> >      {
> >          return $this->logger;
> >      }
> > +
> > +    public function getTemplateData(File $file, $refresh = false)
> > +    {
> > +        //from cache
> > +        if (!$refresh) {
> > +            return $this->importData;
> > +        }
> > +
> > +        $archive = new \ZipArchive();
> > +        $fileName = $file->getBaseName();
> > +        $extractPath = $this->templateDir.$fileName;
> > +
> > +        if (true === $code = $archive->open($file->getPathname())) {
> > +            $res = $archive->extractTo($extractPath);
>
> Sorry, I don't understand. If you remove the templates directory, where
> will the custom templates be stored?
>
> —
> <https://github.com/claroline/Distribution/pull/332/files/29ae0e1d6ac12c5e9a0ac842b1d120b69a4f3cfd#r63867313>
>



-- 
Godfraind Nicolas
5, rue de la Carrière
6592 Monceau-Imbrechies
GSM: 0472 94 80 99

ngodfraind:
Travis is happy now I use the temporary folder. I really think it was a permission issue.

ngodfraind:
LGTM

LaurentGruber:
LGTM

PR #342, opened 2016-05-19
Fix external PR detection in travis

stefk:
This should fix build errors like this one: https://travis-ci.org/claroline/Distribution/builds/131400211#L841 (secured environment variables aren't available in PRs coming from forks).

Not sure about the `after_success` step though.


LaurentGruber:
LGTM

ngodfraind:
LGTM

PR #347, opened 2016-05-19
[ExoBundle] Fix corrections

### LINKED ISSUE #295: Some points on ExoBundle
1 - When there is no question in an exercise, there is a message not translated : 
paper_list_no_paper

2 - When you click on "add a question", before chosing any type of question on the page exercise/admin/question/new/X/X, there is a js error :

Uncaught TypeError: Cannot read property 'valueOf' of undefined 
interactionHole.js:11 

//To find the prototype of wordReponse which is integrated in the prototye of hole
//tableHoles.append('<div id="prototypes" style="display:none"></div>');
var containerWR = container.attr('data-prototype').valueOf();

3 - When a exercise is answered, you can go on the result page. 
You have the user response and the right responses.
When you hover the user response, there is a black infotext.
I suppose it is the feedback. But if there is no feedback this is "null" displayed, which can be most disconcerting if you're not a dev ... Can you hide this infotext when ther is no feedback ?

feedback

Elorfin:
- Remove some debug.
- Partially fix #295 (point 3)
- Do not export the list of questions with the list of papers (data are never used and we can save some DB queries).
- Fix some styles
- Fix feedback info which are not correctly displayed in corrections (when AJAX requests are too long)


LaurentGruber:
LGTM

stefk:
LGTM

PR #351, opened 2016-05-20
Path conditions

CPASimUSante:
Please review this modifications.
With these modifications 
at the path level you can define wether the conditions applies on the next step only or on every step after : 
For a path with  4 steps, and a condition on step 2 : in the first case, the user missing the condition can't go on step 3, in the second case, the user can't access step 3 and 4 (and further if any)

Needs testing with real paths.


stefk:
@CPASimUSante, the weird build errors are probably related to recent changes on the master branch which are not present in your branch. Can you update your PR (there should be an "Update branch" button below)?

CPASimUSante:
ok, i clicked on "update branch" button
Wait and see

plugin/path/Entity/Path/Path.php
```diff
@@ -80,12 +80,22 @@ class Path extends AbstractResource implements PathInterface, \JsonSerializable
     protected $description;
 
     /**
+     * Does the path manage conditions that block all the further steps or just the next one
+     *
+     * @var boolean
+     *
+     * @ORM\Column(name="is_complete_blocking_condition", type="boolean")
+     */
+    protected $isCompleteBlockingCondition;
```
Elorfin: Boolean properties must not start with 'is' or 'has', use `completeBlockingCondition` (you don't need to rename the DB column so you don't have to generate migrations again).

plugin/path/Entity/Path/Path.php
```diff
@@ -282,6 +292,30 @@ public function hasBreadcrumbs()
     }
 
     /**
+     * Get isCompleteBlockingCondition
+     *
+     * @return boolean
+     */
+    public function getIsCompleteBlockingCondition()
```
Elorfin: just name your method `isCompleteBlockingCondition()`, don't need the get (Symfony also finds getters if they start by is or has).

plugin/path/Entity/Path/Path.php
```diff
@@ -282,6 +292,30 @@ public function hasBreadcrumbs()
...
+    }
+
+    /**
+     * Set isCompleteBlockingCondition
+     *
+     * @param boolean $isCompleteBlockingCondition
+     *
+     * @return \Innova\PathBundle\Entity\Path\Path
+     */
+    public function setIsCompleteBlockingCondition($isCompleteBlockingCondition)
```
Elorfin: Becomes `setCompleteBlockingCondition`

plugin/path/Resources/public/js/Path/Controller/PathBaseCtrl.js
```diff
@@ -11,10 +11,12 @@ var PathBaseCtrl = function PathBaseCtrl($window, $route, $routeParams, PathServ
     // Store path to make it available by all UI components
     this.pathService.setId(this.id);
     this.pathService.setPath(this.path);
+    this.pathService.setIsCompleteBlockingCondition(this.path.isCompleteBlockingCondition);
```
Elorfin: same here, remove the `is` from method.

plugin/path/Resources/public/js/Path/Controller/PathBaseCtrl.js
```diff
@@ -11,10 +11,12 @@ var PathBaseCtrl = function PathBaseCtrl($window, $route, $routeParams, PathServ
...
 
     this.currentStep = $routeParams;
 
     this.summaryState = this.pathService.getSummaryState();
+    this.isCompleteBlockingCondition = this.pathService.getIsCompleteBlockingCondition();
```
Elorfin: `this.pathService.isCompleteBlockingCondition()`

plugin/path/Resources/public/js/Path/Service/PathService.js
```diff
@@ -39,6 +39,12 @@
             };
 
             /**
+             * is condition blocking all children of a step
+             * @type {boolean}
+             */
+            var isCompleteBlockingCondition = true;
```
Elorfin: same remark than for PHP : no `is` in property name.

Elorfin:
I've added some little comments about coding standards.
I'm sorry, I will have no time today to check if the code really work as expected : I will give it a try next week.

stefk:
@CPASimUSante, can you update again?

Elorfin:
LGTM

stefk:
LGTM

Issue #337, opened 2016-05-19
[TinyMce] security holes


ngodfraind:
2 security holes:
- Tinymce allows iframe (so you can easily deface a page using them and a bit of css).
- Tinymce doesn't sanitize all javascripts 

```
<script src="xxx"></script> 
```

is allowed so it's not working at all.

LaurentGruber:
users use it to embed service as googleForm ... and they must authenticate to do it soo must we corect it ?

CPASimUSante:
maybe add a checkbox in the general admin : "allow unsecure scripts in editor" ?

PR #355, opened 2016-05-20
Video player upgrade

ngodfraind:
This is not over yet (but it'll be soon enough).

This pr feature:

- global usage of the video player
- almost complete test suite for the video player
- webpack '#' point to distribution-bundle (see #346)
- 'File' field for the angular form
- Lang' field for the angular form
- ngFileUpload for the angular form submission (wich work really well atm)
- video tracks

Note: tracks are stored as Claroline File resources at the moment because it would be nice to be able to pick them with the resource picker later, but it's too much work currently.

Note: The global usage of the video player is debatable but it's the only way to be able to run it in text resources/widgets/... without exploiting tinymce security holes (#337)


ngodfraind:
@claroline/maintainers 

You can review this. It can be squashed before the merge.

main/core/Manager/FileManager.php
```diff
@@ -46,14 +51,55 @@ public function __construct(
         ClaroUtilities $ut,
         ResourceManager $rm,
         StrictDispatcher $dispatcher,
-        $uploadDir
+        $uploadDir,
+        $tokenStorage,
```
stefk: Should be type-hinted.

main/core/Resources/modules/form/Field/File/FileController.js
```diff
@@ -0,0 +1,4 @@
+import FieldController from '../FieldController'
+
+export default class FileController extends FieldController {
+}
```
stefk: Not needed. If `FileController` doesn't add anything to the parent controller, you can use that one directy in `FileDirective`:
```js
this.controller = FieldController
```

main/core/Resources/modules/form/Field/File/FileModelDirective.js
```diff
@@ -0,0 +1,29 @@
+import FileController from './FileController'
```
stefk: Imported but not used?

main/core/Tests/API/Admin/LocaleControllerTest.php
```diff
@@ -0,0 +1,38 @@
...
+     * There is no platform_options.yml test file yet afaik
+     */
+    public function testGetAvailableLanguagesAction()
+    {
+        $admin = $this->createAdmin();
+        $this->login($admin);
+        $this->client->request('GET', '/api/locales/available');
+        $data = $this->client->getResponse()->getContent();
+        $data = json_decode($data, true);
+        $this->assertGreaterThan(2, count($data));
```
stefk: - the `createAdmin` part isn't needed because there's no authentication on that route (at least I didn't see it)
- immediately after the request, there should be an assertion on response code (so that following code doesn't break with weird errors in case of failure)
- the final assertion(s) should probably more precise  and show a sample of what's expected (e.g. with `assertContains`)

plugin/video-player/Resources/modules/track/Controller/TracksModalController.js
```diff
@@ -0,0 +1,65 @@
...
+      resolve: {
+        track: () => {
+          return track},
+        trackForm: () => {
+          return this.trackForm}
+      }
+    })
+
+    modalInstance.result.then(track => {
+      //const data = this.FormBuilderService.formSerialize('roles', panel.panel_facets_role)
```
stefk: Should be removed.

plugin/video-player/Resources/modules/track/Controller/TracksModalController.js
```diff
@@ -0,0 +1,65 @@
+import langs from '#/main/core/Resources/modules/form/Field/Lang/iso.js'
```
stefk: .js extension is superfluous

plugin/video-player/Resources/modules/track/Controller/TracksModalController.js
```diff
@@ -0,0 +1,65 @@
...
+        this.ClarolineAPIService.removeElements([track], this.tracks)
+      }.bind(this),
+      Translator.trans('delete_track', {}, 'platform'),
+      Translator.trans('delete_track_confirm', 'platform')
+    )
+  }
+
+  onEdit (track) {
+    const modalInstance = this.$uibModal.open({
+      template: require('../Partial/edit.html'),
```
stefk: Would be better to use an `import`, so that all dependencies are visible at the top of the module.

plugin/video-player/Resources/modules/track/Directive/TracksButtonDirective.js
```diff
@@ -0,0 +1,12 @@
+import TracksButtonController from '../Controller/TracksButtonController'
+
+export default class TracksButtonDirective {
+  constructor () {
+    this.scope = {}
+    this.restrict = 'E'
+    this.template = require('../Partial/tracks_button.html')
```
stefk: Same.

plugin/video-player/Entity/Track.php
```diff
@@ -0,0 +1,129 @@
...
+use Doctrine\ORM\Mapping as ORM;
+use Claroline\CoreBundle\Entity\Resource\File;
+use Claroline\CoreBundle\Entity\Resource\AbstractResource;
+use JMS\Serializer\Annotation\Groups;
+
+/**
+ * @ORM\Entity(repositoryClass="Claroline\VideoPlayerBundle\Repository\TrackRepository")
+ * @ORM\Table(name="claro_video_track")
+ */
+class Track extends AbstractResource
```
stefk: How does that work? Are tracks visible in the resource manager?

ngodfraind: Not anymore. This can be removed. Good catch.

ngodfraind:
@stefk 

All fixed but the test.

The test one is related to #381. I won't change it in that pr and I'll use that route as the anonymous check for the api in a separate test case.

stefk:
LGTM

ngodfraind:
LGTM

PR #361, opened 2016-05-22
[ExoBundle] Fix copy and import of an Exercise

Elorfin:
Bug fixes : 
- Fix exercise import
- Fix exercise copy

Improvements : 
- refactoring of the ExerciseListener (use cascade for delete, avoid some extra DB queries, remove no longer used events and methods, use manager methods instead of custom implementation)
- use ResourceNode::name instead of the custom property Exercise::title. The custom property as been removed
- simplify ExerciseType form : actually it's just used to create the new Resource (name + published), the other fields was used in the PHP editor which no longer exists.
- some other refactoring in services
- add/correct PHPDocs and comments


LaurentGruber:
LGTM could you review this @UJM-dev ?

Elorfin:
LGTM

PR #338, opened 2016-05-19
API added for the mobile application + merge master

ngydat:

PR #362, opened 2016-05-23
Move preview build to a separate script

stefk:
This should allow to bypass that step effectively for PRs coming from external repositories (see #342).


stefk:
#338 and #351 shoud pass with this. LGTM

ngodfraind:
LGTM

PR #365, opened 2016-05-23
Use new bin dir location

stefk:
Bin directory is now in vendor (composer's default value). See claroline/Claroline#803.


stefk:
Closing in favour of #372

PR #372, opened 2016-05-23
Switch to tar gzip archives

stefk:
Replace zip with tar.gz compression for cache and previews:

- reduce storage and bandwidth use
- preserve symlinks by default

**EDIT**: also use new bin dir location (see #365)


Elorfin:
LGTM

stefk:
LGTM

Issue #140, opened 2016-04-12
ExoBundle : Bug exercise correction


charlotte42:
Dans un exercice si on change les questions et qu'il y a deja des copies, l'affichage de la correction des anciennes copies sont avec les nouvelles questions.

In an exercise when you change questions and which already has copies, the display of the correction of old copies with new questions.

Elorfin:
@marion42 : Can you checkout the branch `exo-improve-papers` and test it ? It's possible that it fixes the problem.

It's still WIP so you may encounters other bugs.

Elorfin:
The code which fix the problem is the methods `PaperManager::exportPaperQuestions(Paper $paper)` and `PaperManager::getPaperQuestions(Paper $paper)`.

marion42:
@Elorfin : Thanks you, but the problem isn't correct. I think the mistake is in PaperManager::exportPaperQuestions(Paper $paper), $question couldn't be find by exercise but by paper.

Elorfin:
@marion42 : Yes the method uses `getOrderQuestion` to get the list of Questions to load. So you only get list of questions linked to the Paper.

```
private function getPaperQuestions(Paper $paper)
    {
        $ids = explode(';', substr($paper->getOrdreQuestion(), 0, -1)); // <-- HERE YOU ONLY GET QUESTIONS LINKED TO THE PAPER

        $questions = [];
        foreach ($ids as $id) {
            $question = $this->om->getRepository('UJMExoBundle:Question')->find($id);
            $questions[] = $question;
        }

        return $questions;
    }
```

Elorfin:
OK, I understand the problem. It's in `PaperManager::getStepsQuestions`, it uses the `StepQuestion` to create the `paper.order` array which is used to display questions in the Player and in the Paper.

marion42:
Yes, I think too, because in `PaperService.js::orderQuestions` don't return all the questions of the paper. And it's base one the paper order.

PR #393, opened 2016-05-24
[ExoBundle] Improve papers

Elorfin:
**Security**
- [x] Check availability of the correction in the API to avoid exporting solution if not needed (security breach)
- [x] Check availability of the score in the API to avoid exporting solution if not needed (security breach)
- [x] Finish check availability of the correction in Angular
- [x] Finish check availability of the score in Angular

**Bug fixes**
- [x] JS error when no generic feedback
- [x] Fix a major bug in correction display (thanks @eldoniel for pointing out and help to debug)
- [x] Fix error 404 on CSS files

**Features**
- [x] Add manual correction needed flags (in the papers list, on the paper, on the questions of the paper)

**Improvements**
- [x] Optimize step navigation. Due to the AngularJS router each steps called the API to retrieve the current Paper. The current attempt is now cached in Angular (in `UserPaperService`).
- [x] Remove User name column in papers list for regular Users.


plugin/exo/Controller/Api/ExerciseController.php
```diff
@@ -225,17 +224,27 @@ public function countFinishedPaperAction(User $user, Exercise $exercise)
      * associated with the exercise.
      *
      * @EXT\Route("/exercises/{exerciseId}/papers/{paperId}", name="exercise_paper")
+     * @EXT\ParamConverter("user", options={"authenticatedUser" = false})
```
stefk: I'm not sure that converter will handle anonymous requests correctly. Its code is actually weird and bloated. If you simply want to get the current user and return a 403 otherwise, [this one](https://github.com/claroline/Distribution/blob/master/main/core/Converter/CurrentUserConverter.php) is a better fit. Example [here](https://github.com/claroline/Distribution/blob/master/plugin/exo/Controller/Api/ExerciseController.php#L113).

plugin/exo/Manager/PaperManager.php
```diff
@@ -459,23 +430,36 @@ private function exportPaperQuestions(Paper $paper)
...
-        $qIds = explode(';', substr($paper->getOrdreQuestion(), 0, -1));
+        $ids = explode(';', substr($paper->getOrdreQuestion(), 0, -1));
 
-        //to keep the questions order
         $questions = [];
-        foreach ($qIds as $qid) {
-            $q = $this->om->getRepository('UJMExoBundle:Question')->find($qid);
-            $questions[] = $q;
+        foreach ($ids as $id) {
+            $question = $this->om->getRepository('UJMExoBundle:Question')->find($id);
```
stefk: Not related to this PR, but we must absolutely do something about this (one query instead of a loop of queries).

plugin/exo/Controller/Api/ExerciseController.php
Elorfin: Yes I have seen the one you point out too. 

In fact I use the current one because it lets pass the anonymous users. So if the user is logged I get a `User` instance and if anonymous, I get the string `anon.`.

This will permits to anonymous users to access and play Exercises. The permission check will let the user pass or not depends on the Resource access config.

But you are right in this place anonymous should not have access (or they will have access to all anonymous papers). Also the type hinting of the function is incorrect. I will fix it.

I can't remember where I have find usages of the converter but there are some features using it.
I wanted to use it in other actions that need anonymous access. Do you advise not to ?

plugin/exo/Manager/PaperManager.php
Elorfin: Yes. But this whole function needs to be changed because it causes issue when there are papers with Questions which are no longer into the Exercise (#140).

Also there is another "query loop" lower in the function for StepQuestions.

Elorfin:
I have removed the possible buggy control access on papers.

@claroline/maintainers : can we merge ?

stefk:
@Elorfin, do you want to rebase/squash/rewrite your commits or do we merge as is?

Elorfin:
Nope, there are many features in this PR and I don't want to loose history. Even if I have some ugly merge and fix commits, every other commits is (relatively) independent.

stefk:
Ok, no problem. Don't forget you can use rebase and rebase -i locally for a complete control over your history.

stefk:
LGTM

purplefish32:
LGTM

Issue #394, opened 2016-05-24
[Travis] PHP CS ends with an error


Elorfin:
The PHP CS step in the Travis build generates an error that makes the build fail ([Failing build](https://travis-ci.org/claroline/Distribution/builds/132644637)).

The exception : 
```
The command "$ROOT/vendor/bin/phpmd --exclude main/core/Resources/bundle-creator,main/web-installer `cat git_diff_files.txt | xargs | sed -e :a -e '$!N; s/ /,/; ta'` text phpmd.xml" exited with 1.

$ $ROOT/vendor/bin/php-cs-fixer fix --dry-run --diff --config-file=main/dev/Resources/travis/php-cs.php

Loaded config from "main/dev/Resources/travis/php-cs.php"

[ErrorException]                               

file_get_contents(): Filename cannot be empty 
```

Maybe @purplefish32 or @stefk have an idea ?

Also I'm not sure about the labels to add on this kind of problems.

stefk:
@Elorfin, I saw that error a couple of times. I think the `git diff-tree` command used in the travis script to list modified files has a wrong or missing option. For some reason, when a file has been deleted on the master branch and the PR branch isn't up-to-date, that file ends up in the diff.  That's what is making php-cs-fixer crash later.

 I've made a clone of the current state of your branch, I'll try to debug the issue tomorrow. In the meantime, if you can't wait for a fix, I'm pretty sure you could get rid of the error by merging or rebasing your branch on the master.

Elorfin:
Not really urgent, it is mainly to log the issue.
You are right, an update from the `master` and all works fine now.

Thanks for the help.

PR #395, opened 2016-05-25
[Test only]Travis fail demo #394

stefk:
Just a test to reproduce and debug #394. Not supposed to be merged.


stefk:
Should be fixed in #396.

PR #399, opened 2016-05-25
Facets PR failure test

stefk:
#396 leftover test. Not supposed to be merged.


stefk:
Should be fixed in #402.

PR #400, opened 2016-05-25
[CoreBundle] Allow anonymous Users while using CurrentUserConverter

Elorfin:
- [x] Update the converter
- [x] Add Tests for the converter
- [x] Replace usage of `AuthenticatedUserConverter` which are buggy ( with `"authenticatedUser"=false`)

In `WorkspaceController`, I have removed access to anonymous to the action `removeUserFromQueue`. Because the user is used to make a query so I think it's required. can you confirm @ngodfraind ?

We will need to gradually remove the use of the `AuthenticatedUserConverter` (411 usages found) as the extra parameters for message and translations seems to never be used.


main/core/Controller/WorkspaceController.php
```diff
@@ -284,7 +280,7 @@ public function listWorkspacesWithSelfRegistrationAction($search = '')
      *     defaults={"page"=1},
      *     options={"expose"=true}
      * )
-     * @EXT\ParamConverter("currentUser", options={"authenticatedUser" = true})
```
stefk: Be careful: with `authenticatedUser` set to true, the old converter actually redirects to the login form. That may be the expected behaviour here. cc @ngodfraind

main/core/Converter/CurrentUserConverter.php
```diff
@@ -65,6 +83,15 @@ public function apply(Request $request, ParamConverter $configuration)
...
     public function supports(ParamConverter $configuration)
     {
-        return $configuration->getConverter() === 'current_user';
+        if (!$configuration instanceof ParamConverter) {
+            return false;
+        }
+
+        $options = $configuration->getOptions();
+        if (isset($options['allowAnonymous']) && !is_bool($options['allowAnonymous'])) {
+            return false;
```
stefk: Having a non-boolean value for `allowAnonymous` is a configuration error IMO. Returning false means the error is silenced and the converter completely bypassed. An exception (here or in `apply`) would make more sense.

main/core/Tests/Unit/Converter/CurrentUserConverterTest.php
```diff
@@ -0,0 +1,106 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Claroline\CoreBundle\Converter;
+
+use Symfony\Component\HttpFoundation\ParameterBag;
+use Claroline\CoreBundle\Library\Testing\MockeryTestCase;
+use Claroline\CoreBundle\Entity\User;
+
+class CurrentUserConverterTest extends MockeryTestCase
```
stefk: Tests of the core bundle have to be included explicitely in the test suite (because there's a lot of old/broken tests in that bundle). See https://github.com/claroline/Distribution/blob/master/phpunit.xml#L30

plugin/social-media/Controller/ShareActionController.php
```diff
@@ -25,7 +25,7 @@ class ShareActionController extends Controller
 {
     /**
      * @Route("/share/form/{resourceId}", name="icap_socialmedia_share_form", )
-     * @ParamConverter("user", options={"authenticatedUser" = true})
+     * @ParamConverter("user", converter="current_user")
```
stefk: Same remark than above (login form redirection). cc @ptsavdar

Elorfin:
The tests have been added to the PHPUnit test suite.
I have also modified the parameters checks (thanks @stefk).

I wait for @ngodfraind and @ptsavdar advices for what to do with the too controller actions.

stefk:
ping @ngodfraind, @ptsavdar

ngodfraind:
@kitan1982 did this part but it looks like you're right.

LaurentGruber:
LGTM

stefk:
@LaurentGruber, there are still unresolved issues here.

plugin/social-media/Controller/ShareActionController.php
LaurentGruber: J'ai tester l'accès au partage sans être connecté, il renvoi vers la page de login et connecté tout se passe bien, donc ça me semble ok.
LGTM

ptsavdar:
LGTM

LaurentGruber:
LGTM

Issue #324, opened 2016-05-18
Bug in Exo Bundle, open question (distribution bundle)


CPASimUSante:
I created an exercise with 2 open questions (short answer). When i answer this excercise, if i leave a field blank, and i click on "finish exercise", nothing happens, no message. But in the JS console i have

`PUT http://claro-distrib/app_dev.php/exercise/api/papers/2/steps/1               422 (Unprocessable Entity)`
And in the js/network tab : 
`["Answer data must be a string, NULL given"]`

Elorfin:
OK.
I will take a look, We need to register an empty string instead of null if the user does not answer the Question.

I need to check for the other question types too.

PR #406, opened 2016-05-26
Exo fix 324

eldoniel:
This fixes the following issue:
https://github.com/claroline/Distribution/issues/324

The holes with multiple propositions in cloze questions were also concerned by this bug. This is fixed as well.


Elorfin:
LGTM

stefk:
LGTM

Issue #421, opened 2016-05-26
Import QTI with xml with the same name


david42:
only one question is imported

PR #422, opened 2016-05-26
PR for issue #421

david42:
only one question is imported if the zip file contains several xml with the same name


Elorfin:
LGTM

LaurentGruber:
LGTM

PR #425, opened 2016-05-26
Revert "[ExoBundle] Improve papers"

stefk:
Reverts claroline/Distribution#393


stefk:
So that we can reapply it without losing history.

purplefish32:
LGTM

stefk:
LGTM

PR #429, opened 2016-05-27
Exo qti export

charlotte42:
Miss that importing a exercice


plugin/exo/Controller/QtiController.php
```diff
@@ -215,47 +179,46 @@ public function ExportAction($id)
...
+        $qtiSer = $this->container->get('ujm.exo_qti');
+        $qtiRepo = $this->container->get('ujm.exo_qti_repository');
+
+        foreach ($exercise->getSteps() as $step) {
+            $questions = $questionRepo->findByStep($step);
+
+            $qtiSer->createQuestionsDirectory($questions, $step->getOrder());
+            $dirs = $qtiSer->sortPathOfQuestions($qtiRepo, $step->getOrder());
+            $i = 'a';
+            var_dump($dirs);
```
stefk: `var_dump` should be removed.

plugin/exo/Services/classes/QTI/QtiServices.php
```diff
@@ -48,8 +59,34 @@ public function sortPathOfQuestions(QtiRepository $qtiRepo, $stepDir = '')
...
+     * create the directory questions to export an exercise and export the qti files.
+     *
+     * @param \UJM\ExoBundle\Entity\Question[] $questions
+     * @param int                              $numStep
+     */
+    public function createQuestionsDirectory(array $questions, $numStep)
+    {
+        $qtiRepo = $this->container->get('ujm.exo_qti_repository');
+
+        @mkdir($qtiRepo->getUserDir().'questions');
```
stefk: Why do you use the silence `@` operator?

plugin/exo/Controller/QtiController.php
charlotte42: .... sorry

plugin/exo/Controller/ExerciseController.php
```diff
@@ -344,13 +344,16 @@ public function importValidateAction()
...
         $this->assertHasPermission('ADMINISTRATE', $exercise);
+        $em = $this->getDoctrine()->getManager();
 
         $step = new Step();
         $step->setText(' ');
         $step->setNbQuestion('0');
         $step->setDuration(0);
         $step->setMaxAttempts(5);
-        $step->setOrder($exercise->getSteps()->count() + 1);
+        $result = $em->getRepository('UJMExoBundle:Step')->getMaxOrder($exercise);
```
Elorfin: You don"t need a DB query here. You just have to count Steps from Exercise (`$exercise->getSteps()->count() + 1`).

So you can remove your custom Repository method `getMaxOrder`.

plugin/exo/Resources/config/services.yml
```diff
@@ -154,6 +154,7 @@ services:
 
     ujm.exo_qti:
         class: UJM\ExoBundle\Services\classes\QTI\QtiServices
+        arguments: [@service_container]
```
Elorfin: Bad practice to inject the whole `Container` into a Service. Juste inject services you really use.

plugin/exo/Services/classes/QTI/QtiServices.php
ngodfraind: It's usually done to avoid crashes if the directory already exists when no check is done.

Elorfin: Isn't that better to actually do the check instead of just silence errors ?

stefk: IMO every call to `mkdir` should be removed. If you use Symfony's `Filesystem` component, you get proper exceptions in case of failures, and it doesn't complain if the directory already exists.

Elorfin: +1 for using symfony `Filesystem`.

plugin/exo/Controller/ExerciseController.php
charlotte42: Je l'ai remplacé car si on supprime une step (pas la derniére) et que l'on en crée aprés une nouvelle l'order ne sera pas le bon. ex : 1 2 3 je supprime la 2 il va en compter 2 donc j'aurais deux fois 3. Aprés si ça ne va pas je peux changer.

plugin/exo/Services/classes/QTI/QtiServices.php
charlotte42: J'ai réutilisé ce qu'il y avait déjà mais si c'est mieux de faire un test pour savoir si il existe je change pas de soucis.

plugin/exo/Controller/ExerciseController.php
Elorfin: OK, but it's not the correct way to fix it. 
We need to keep the order correct in DB. So we need to update step orders when deleting a Step and not dealing with incorrect data.

plugin/exo/Resources/config/services.yml
charlotte42: pareil que le @ j'ai fait comme c'était dans d'autre service, je vais modifier

plugin/exo/Controller/ExerciseController.php
Elorfin: [#433](https://github.com/claroline/Distribution/pull/433/files#diff-413883e78d14ce9439e0d2c33e634bb8R78).

I have fixed it. (Anyway the method you try to fixed has been refactored in the PR I have linked)

plugin/exo/Services/classes/QTI/QtiServices.php
stefk: @charlotte42, pas obligé, mais si tu veux le faire, le plus simple/propre, c'est : 
```php
use Symfony\Component\Filesystem\Filesystem;

$fs = new Filesystem();

$fs->mkdir($qtiRepo->getUserDir().'questions');
// ...
$fs->mkdir($qtiRepo->getUserDir().'questions/'.$numStep.'/'.$numStep.'_question_'.$i);
```

plugin/exo/Controller/ExerciseController.php
charlotte42: Ok je ne savait pas que tu t'en occupais

Elorfin: I have updated the method in 2min when you have explained why you wanted to do that.

Also I fix issue directly when I find them ;).

Elorfin: So please, restore the old code (to not have conflict with my PR) and remove your new repository method which is not needed.

LaurentGruber:
@UJM-dev vous voulez bien régler les conflits de merge ?

LaurentGruber:
LGTM

Elorfin:
LGTM

PR #430, opened 2016-05-27
Add mobile app API (update #338)

ngodfraind:


ngodfraind:
ngydat doesn't work here anymore so I took his work, merged the master and run php-cs-fixer

related to #338

plugin/exo/Tests/Controller/Api/ExerciseControllerChoiceTest.php
```diff
@@ -63,6 +63,7 @@ public function testSubmitAnswerInInvalidFormat()
     {
         $pa1 = $this->persist->paper($this->john, $this->ex1);
         $this->om->flush();
+        //$this->logIn($this->john);
```
ngodfraind: comments can be removed

ngodfraind:
@claroline/maintainers 

Beside the superflous comment in the test, LGTM.

stefk:
@ngodfraind, I don't see any security checks in the controllers.

ngodfraind:
ClientController is supposed to be public and the other requires the user to be connected.
We could throw an access denied if the ROLE_USER is not present to avoid crashes.

stefk:
Ok, so there're two things left:
- fix indentation of the last method of the first controller
- remove tokenStorage in the second one, and use the `current_user` converter (that way actions will become one-liners)

ngodfraind:
I did more digging and realised the first controller doesn't return always the good client.
It should retrieve a client whose grant type is only password and refresh. Otherwise security will be an issue.

ngodfraind:
The API is not tested either wich is an issue.

LaurentGruber:
LGTM

stefk:
@LaurentGruber, it can be merged as is, see https://github.com/claroline/Distribution/pull/430#issuecomment-222161616.

ngodfraind:
It's only missing some plugins tests now.
Note that I changed some API urls so they'll need to be changed aswell in the mobile application.

ngodfraind:
I'll fix what's broken this afternoon.

ngodfraind:
Tests are only partially done but it's mainly because persisting the correct stuff in the database is hard currently. We can improve them later.

main/core/Controller/API/Admin/ClientController.php
```diff
@@ -0,0 +1,59 @@
...
+   {
+       $this->oauthManager = $oauthManager;
+   }
+
+    /**
+     * Get the client id and the client secret.
+     *
+     * @Route("/client/public", name="claro_id_secret", defaults={"_format":"json"})
+     */
+    public function getIdsecretAction()
```
stefk: Naive notes/questions because I don't really know oauth stuff but:
- why would a public api return a client secret?
- the controller is said to return an id/secret but it calls a `findUsername` method, so what is it really doing? 
- `findUsernameClient` logic seems almost random

ngodfraind:
- api & secret are both required to setup an oauth authentication and you must use them in you authentication http request.
- username is an oauth grant type allowing log a user. Therefore clients whose grant type are username (+ refresh) can become public. Currently, it'll just return the 1st client it finds. We could add a "public" flag to the differents clients but I'd rather do that when I'll rewrite the oauth system with angular.

stefk:
LGTM

ngodfraind:
LGTM

Issue #104, opened 2016-04-07
ExoBundle : conflict between Graphic and Match questions


purplefish32:
_From @Elorfin on March 31, 2016 7:34_

In the Player, if there are a Graphic question and a Match question into the same step there are conflicts in the JS and it doesn't work properly.

_Copied from original issue: claroline/ClaroSupport#22_

purplefish32:
@Elorfin @eldoniel : is this issue still valid or can we close it ?

PR #440, opened 2016-05-29
[ExoBundle] Graphic update and fixes

Elorfin:
This PR contains a new module for the Timer and various little fixes.

The big part is the new implementation of the Graphic question.

This implementation fixes the following bugs : 
- conflict between events (with other graphic and match). This fixes in part the #104.
- fix bug when displaying graphic correction
- fix possible misplacement of the points if a parent element is set to `relative`. It causes the bug above and possible problem in Path (but not tested for path).
- follow the json-quizz schema to transfer data (easier to process data in angular than the previous implementation)

The new implementation include the following changes :
- no dependency to jquery UI (only jquery) 
- the image is now responsive (we use the original size to save answer). We will need to add a magnify option, but no time now. This permits to avoid the scroll while drag n drop which is pretty annoying.
- uses a canvas to draw the answer zones (will be the only way we can display "poly" zone in the future).
- instead of pre-append cursors to the view when no response, the cursor are added by clicking on the image (make calculation easier and more robust).

This has been tested in FF, Chrome and IE.

There are still issues with immediate feedback (cursor img is not replaced by validation icon) and all answer zones are displayed (instead of only those with a cursor in it). But the current implementation seems to fix more problem than adding ones.


stefk:
LGTM

ngodfraind:
LGTM

Issue #278, opened 2016-05-13
[CoreBundle] Create Angular utilities to manage tinyMCE


Elorfin:
Add a TinymceConfigService and trustAsHtml filter in the CoreBundle.
- The config service will permit to always have the same config for all TinyMCE fields.
- We need to regularly redefine trustAsHtml (we have the same filter in ExoBundle).

For now, PathBundle / ExoBundle and WebsiteBundle have their own implementation.

@ptsavdar : to respond to what you say on #277, we can use the ES6 compiled code in assetic and so mix ES6 / ES5. 

I remember, we have done that with @ngodfraind when he has shown me how I can use ES6 + webpack for PathBundle. That was a try so only one of the module is in ES6 and the others in ES5 and it doesn't seem to be a problem.

Someone know issues when mixing ES5 and ES5 compiled from ES6 ?

stefk:
I agree, we can use ES6 compiled code in ES5 code. The only thing we must take care of is the size/content of compiled modules. AFAIK all the modules we've built with webpack are completely self-contained (i.e. include the module code + its dependencies: angular, angular-ui, etc.). Obviously if we want to reuse modules in non es6/webpack code, we must make them lighter. But that's a kind of webpack config tweaking we must do anyway.

CPASimUSante:
I'm not sure what sort of config you're talking about : only main general configuration (just the init), or complete configuration with the list of plugins and toolbars ? 
For the latter, i think this part should be kept separated or at least be overridable :
- Some bundle could require less (or more plugins)
- some Claroline instances could be more or less permissive for the editing tools (for instance prevent using text or background color feature...)

So, centralizing the Tinymce code is necessary, but i think it should not be done at the expense of customization.

PR #441, opened 2016-05-29
[ExoBundle] Step metadata

Elorfin:
* Export metadata of the step in a `meta` property (like the Exercise)
* Display a modal in step list to edit step metadata

For now only the parameters `description` and `maxAttempts` are managed as the others are not implemented.

@LaurentGruber : you can simulate a step with only `Content` items by adding a description HTML to step without adding questions.

### DIFF
--- a/plugin/exo/Resources/public/js/angular/Common/Services/TinyMceService.js
+++ b/plugin/exo/Resources/public/js/angular/Common/Services/TinyMceService.js
@@ -1,0 +1,47 @@
+var TinyMceService = function TinyMceService() {
+
+};
+
+// Set up dependency injection
+TinyMceService.$inject = [];
+
+TinyMceService.prototype.getConfig = function getConfig() {
+    var config = {};
+
+    var tinymce = window.tinymce;
+    tinymce.claroline.init    = tinymce.claroline.init || {};
+    tinymce.claroline.plugins = tinymce.claroline.plugins || {};
+
+    var plugins = [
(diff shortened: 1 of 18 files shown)


plugin/exo/Resources/public/js/angular/Common/Services/TinyMceService.js
```diff
@@ -0,0 +1,47 @@
+var TinyMceService = function TinyMceService() {
```
ngodfraind: I think it should be located somewhere in the core bundle as it's pretty usefull everywhere.

Elorfin: Yes, we are all OK with that. But I did not want to add angular module in ES5 in the core.
And as discussed with @ptsavdar and @stefk it seems that if I create an ES6 module and use webpack the generated file will contain all the dependencies needed (like angular). I wanted to avoid this overload.

LaurentGruber:
LGTM

plugin/exo/Resources/public/js/angular/Common/Services/TinyMceService.js
stefk: @ngodfraind, see #278. Scheduled for 16.9.

Elorfin: so are you ok to merge it as is ?

stefk:
LGTM

PR #443, opened 2016-05-30
[CoreBundle] Allows to edit maintenance message

### LINKED ISSUE #56: Message personnalisé du mode maintenance
_From @clarophil on March 23, 2016 14:57_

Il serait intéressant de pouvoir personnaliser dans l'admin le message qui s'affiche quand on est dans le mode maintenance par exemple pour dire combien de temps durera approximativement la maintenance.

_Copied from original issue: claroline/Claroline#732_

kitan1982:
-  Allows to edit maintenance message. 
-  Closes #56

### DIFF
--- a/main/core/Controller/Administration/ParametersController.php
+++ b/main/core/Controller/Administration/ParametersController.php
@@ -11,6 +11,7 @@
 
 namespace Claroline\CoreBundle\Controller\Administration;
 
+use Claroline\CoreBundle\Entity\Content;
 use Claroline\CoreBundle\Entity\SecurityToken;
 use Claroline\CoreBundle\Event\StrictDispatcher;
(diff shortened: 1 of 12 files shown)


stefk:
Is it really a good idea to store that information in db? If the database isn't available at some point (which isn't unlikely in a maintenance operation), the script will crash completely. Another thing is that it forces us to do weird/weak stuff like [this](https://github.com/kitan1982/Claroline/blob/be3b1beffc5d5b4bc474396544f05d0fec7cd6b4/web/maintenance.php#L48), whereas a simple file (.php, .ini, whatever) would do the trick perfectly.

kitan1982:
The script doesn't crash if the database isn't available. It will only render an empty message.
But it is sure that the code to retrieve the database informations is rather ugly.
I will store the message in a php file.

PR #445, opened 2016-05-30
Exo generic feedbacks fix

eldoniel:
This fixes a bug in the correction: generic feedbacks were shown in the correction page (at the end of the exercise), while it should have been shown only while playing the exercise itself.


plugin/exo/Resources/public/js/angular/Exercise/Services/FeedbackService.js
```diff
@@ -32,11 +32,11 @@ FeedbackService.prototype.config = {
     visible: false,
     
     /**
-     * The state of the feedback
+     * The state of the feedback for each question
      * (0 : all found, 1 : partially found, 2 : error)
      * @type {integer}
      */
-    state: -1
+    state: []
```
Elorfin: you will need to initialize a new object here instead of array.
this will avoid to have `undefined` values if the IDs of the questions are not sequential.
also if we use id that are not numbers, this will not work as expected.

plugin/exo/Resources/public/js/angular/Question/Controllers/Type/ChoiceQuestionCtrl.js
```diff
@@ -140,7 +140,7 @@ ChoiceQuestionCtrl.prototype.onFeedbackShow = function onFeedbackShow() {
         }
     }
     
-    this.feedbackState = this.ChoiceQuestionService.answersAllFound(this.question, this.answer);
+    //this.feedbackState = this.ChoiceQuestionService.answersAllFound(this.question, this.answer);
```
Elorfin: commented code should not be commited

Elorfin:
LGTM

stefk:
LGTM

PR #465, opened 2016-05-30
Exo generic feedbacks fix

eldoniel:
This fixes the following bug: when having multiple match questions of the same type (all bind or all drag) in a step, there was an exception when we tried to save an answer.


LaurentGruber:
LGTM

Elorfin:
LGTM

PR #467, opened 2016-05-31
[CoreBundle] Administration user list role management.

### LINKED ISSUE #88: Improvement : Add multiple roles in administration in user list
_From @LaurentGruber on March 23, 2016 12:35_

_From @CPASimUSante on October 26, 2015 17:23_

As an administrator, I've created several users, but i forgot to create a specific role for them beforehand. So I create the role and then i want to give it to the users : i have to edit each one to add the role. There is a checkbox to select several user, but no button to assign role(s). It would be useful i think, as a platform administrator.

_Copied from original issue: claroline/Claroline#712_

_Copied from original issue: claroline/ClaroSupport#6_

ngodfraind:
Fix #88


main/core/Controller/API/User/RoleController.php
```diff
@@ -0,0 +1,63 @@
...
+        return $this->authorization->isGranted('ROLE_ADMIN');
+    }
+
+    private function throwsExceptionIfNotAdmin()
+    {
+        if (!$this->isAdmin()) {
+            throw new AccessDeniedException('This action can only be done by the administrator');
+        }
+    }
+}
```
stefk: A simple `@SEC\PreAuthorize("hasRole('ROLE_ADMIN')")` would be enough.

ngodfraind: Can I use it a the beginning of a method ? This won't be the only method of this controller later and the admin won't be the only one to have accesses.

main/core/Tests/API/User/UserControllerTest.php
```diff
@@ -23,7 +23,7 @@ protected function setUp()
         parent::setUp();
         $this->persister = $this->client->getContainer()->get('claroline.library.testing.persister');
     }
-
+/*
```
ngodfraind: uncomment tests

main/core/Controller/API/User/RoleController.php
stefk: Yep: http://jmsyst.com/bundles/JMSSecurityExtraBundle/master/annotations

main/core/Resources/modules/users/Controller/UserController.js
```diff
@@ -161,6 +197,31 @@ export default class UserController {
...
+            users +=  this.selected[i].username
+            if (i < this.selected.length - 1) users += ', '
+        }
+
+        let roles = ''
+
+        for (let i = 0; i < this.roles.length; i++) {
+            roles +=  this.translate(this.roles[i].translation_key)
+            if (i < this.roles.length - 1) roles += ', '
+        }
```
stefk: You should avoid low-level array/string manipulation like that. These 12 lines can be reduced to 2:
```js
const users = this.selected.map(s => s.username).join(', ')
const roles = this.roles.map(r => this.translate(r.translation_key)).join(', ')
```

ngodfraind: nice

ngodfraind:
Is it possible to know what files php-cs-fixer found ? Because it doesn't seem to be in the main/core directory.

ngodfraind:
@stefk why didn't travis start ?

stefk:
@ngodfraind, IDK, sometimes the notification gets lost. Just do:
```` 
git commit --amend
git push --force
```
(you can keep the original commit message)

ngodfraind:
LGTM

stefk:
LGTM

PR #472, opened 2016-05-31
[SurveyBundle] Disables answered question edition

### LINKED ISSUE #380: [SurveyBundle] Answer lost if the question is modified
I create the survey with one question and then publish it
A student fills the survey
Then I modify the question and all results are lost but the number of answer is still correct.

I think that the modification should be forbidden with the survey is publised (as exercize), there is always to copy the survey and then modify it.

image

kitan1982:
- Disables question edition when there are answers associated to it.
- closes #380


LaurentGruber:
LGTM

Elorfin:
LGTM

LaurentGruber:
LGTM

PR #477, opened 2016-05-31
[CoreBundle] Fixing web installer

ngodfraind:
- memory requirement to 512M (required during the installation assetic:dump)
- same code base for installation everywhere (runs from the previous-installed.json; must be empty for the install) as the current install from the web installer crashes.
- removing dead test


Elorfin:
LGTM

LaurentGruber:
LGTM

PR #483, opened 2016-05-31
[CoreBundle] Registers last view mode of resources manager

### LINKED ISSUE #207: Gestionnaire de ressource : vue liste
Mémoriser la demande d'une vue (liste ou icone) et afficher cette vue par défaut lors du retour dans l'espace de ressource

kitan1982:
- Registers last view mode of resources manager
- closes #207

### DIFF
--- a/main/core/Controller/ResourceController.php
+++ b/main/core/Controller/ResourceController.php
@@ -11,6 +11,7 @@
 
 namespace Claroline\CoreBundle\Controller;
 
+use Claroline\CoreBundle\Manager\UserManager;
 use Exception;
 use Symfony\Bundle\TwigBundle\TwigEngine;
 use Symfony\Component\Translation\TranslatorInterface;
@@ -57,6 +58,7 @@ class ResourceController
     private $fileManager;
     private $transferManager;
     private $formFactory;
+    private $userManager;
 
     /**
      * @DI\InjectParams({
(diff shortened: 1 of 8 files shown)


main/core/Controller/ResourceController.php
```diff
@@ -1135,6 +1140,25 @@ public function importAction(ResourceNode $directory)
...
+     *     name="claro_resource_manager_display_mode_register",
+     *     options={"expose"=true}
+     * )
+     * @EXT\ParamConverter("user", converter="current_user", options={"allowAnonymous"=true})
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function resourceManagerDisplayModeRegisterAction(User $user, $index, $displayMode)
+    {
+        if ($user !== 'anon.') {
```
stefk: That condition will always be true due to the type hinting above.

stefk: Correct type hinting should probably be: `User $user = null`

stefk: (and the current_user converter doesn't return '.anon')

main/core/Controller/ResourceController.php
```diff
@@ -1135,6 +1140,25 @@ public function importAction(ResourceNode $directory)
...
+            $this->userManager->registerResourceManagerDisplayModeByUser($user, $index, $displayMode);
+        }
+
+        return new Response(200);
+    }
+
     public function deleteNodeConfirmAction(ResourceNode $node)
     {
         throw new \Exception('hey');
```
stefk: Not your code but seems like garbage.

main/core/Resources/public/js/resource/manager/views/resource/actions.js
```diff
@@ -335,7 +337,7 @@
                 && this.isReadyToPaste
                 && (!this.isCutMode || this.checkedNodes.directoryId !== event.id);
 
-            var listViewActivated = this.displayMode === 'default' ? false: true;
+            var listViewActivated = this.displayMode === 'list' ? true: false;
```
stefk: Should be simplified to:

```php
var listViewActivated = this.displayMode === 'list';
```

LaurentGruber:
LGTM

kitan1982:
LGTM

PR #487, opened 2016-05-31
[CoreBundle] nelmio-api-doc as a dev dependency so it doesn't crash d…

ngodfraind:
…e prod build.


stefk:
- is it really a dev dependency? I found some @Api annotations in the code, I don't know if it's related
- why does it crash prod builds?

ngodfraind:
Nope. Forgot the annotations.
Now I fixed its version on ~2.0 instead of dev-master.

It crashed because it can't find the last commit.

PR #491, opened 2016-05-31
[ExoBundle] closes #140

### LINKED ISSUE #140: ExoBundle : Bug exercise correction

Elorfin:
I think it can be improved and it requires further tests if you want to continue an attempt with a deleted question.

But at least the corrections and scores are  OK. When loading the list of papers, I also load the set of questions for each paper. Can be improved to avoid performance impact...

I create a "fake" step at the end of the step list to get the questions which are no longer in the Exercise. So in that step, the questions are ordered. But the step is not at the correct place and the questions maybe not belong to the same step (this information has been lost when deleting).

### DIFF
--- a/plugin/exo/Controller/Api/ExerciseController.php
+++ b/plugin/exo/Controller/Api/ExerciseController.php
@@ -152,7 +152,7 @@ class ExerciseController
             return new JsonResponse($this->paperManager->exportExercisePapers($exercise));
         }
 
-        return new JsonResponse($this->paperManager->exportUserPapers($exercise, $user));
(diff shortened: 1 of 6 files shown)


purplefish32:
LGTM

ngodfraind:
LGTM

Elorfin:
LGTM

ngodfraind:
LGTM

PR #495, opened 2016-05-31
[CoreBundle] Fixing translation (removing space).

ngodfraind:


ngodfraind:
There is no double check here ? I can already merge.

PR #497, opened 2016-06-01
Fixing update

ngodfraind:
Sometimes, a bundle is present in the sources but was set to false in bundles.ini or not installed. This PR fix this case during an update from <7.x


ngodfraind:
@stefk Travis crashed: ``` fatal: ambiguous argument '7.x': unknown revision or path not in the working tree. ```

PR #498, opened 2016-06-01
[CoreBundle] Public profile

ngodfraind:
The profile_view route doesn't exist anymore. This is a q&d fix. Will do something better afterwards (like removing every profile_view route occurences).

### DIFF
--- a/main/core/Controller/ProfileController.php
+++ b/main/core/Controller/ProfileController.php
@@ -136,6 +136,20 @@ class ProfileController extends Controller
         }
     }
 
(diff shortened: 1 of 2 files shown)


ngodfraind:
@stefk could it crashes because it reference the d20c1a2 commit ?

stefk:
@ngodfraind, it looks like you created the PR branch from the master, not from the 7.x branch. Same for #497.

ngodfraind:
is it why it crashes ?

PR #525, opened 2016-06-02
[VideoPlayer] Allows videojs to be used in path resources

ngodfraind:
This is pretty dirty atm (because it triggers an infinite loop if you do it like you would expect) so a better solution is welcome if you can find one.

Also jquery could be removed but it's usefull for small scripts like this one.

### DIFF
--- a/plugin/video-player/Resources/js/observe.js
+++ b/plugin/video-player/Resources/js/observe.js
@@ -3,29 +3,30 @@ export default function observe (selector, callback, containers = [document.body
 
   const initialized = []
   // create an observer instance
+
   const observer = new MutationObserver(mutations => {
     mutations.forEach(mutation => {
-      if (mutation.type == 'attributes') {
-          const toInit = $(mutation.target).find(selector)
-          toInit.each((i, el) => {
-              let keepGoing = true
-              initialized.forEach(id => {
-                  //this is required because otherwise videoJs goes into infinite loop for unknown reason
-                  //as it automatically adds html_5_api at the end of an id wich triggers the observer wich trigger video js
-                  //to add html_5_api to the id wich trigger the observer... etc.
-                  if (el.id.indexOf(id) === 0) keepGoing = false
-              })
-
-              if (keepGoing) {
-                  callback(el)
-                  initialized.push(el.id)
-              }
+      if (mutation.type == 'childList') {
+        $(mutation.addedNodes).find(selector).each((i, el) => {
+          let keepGoing = true
+          initialized.forEach(id => {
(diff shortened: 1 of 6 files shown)


LaurentGruber:
I dont have bette solution anyone else @claroline/maintainers ?

plugin/video-player/Resources/js/observe.js
```diff
@@ -0,0 +1,31 @@
...
+    mutations.forEach(mutation => {
+      if (mutation.type == 'attributes') {
+          const toInit = $(mutation.target).find(selector)
+          toInit.each((i, el) => {
+              let keepGoing = true
+              initialized.forEach(id => {
+                  //this is required because otherwise videoJs goes into infinite loop for unknown reason
+                  //as it automatically adds html_5_api at the end of an id wich triggers the observer wich trigger video js
+                  //to add html_5_api to the id wich trigger the observer... etc.
+                  if (el.id.indexOf(id) === 0) keepGoing = false
```
Elorfin: Missing `{}` for the condition.

Elorfin:
Sorry never played with `MutationObserver` and videoJS.
Just found a little CS typo ;).

stefk:
I don't think it should be merged as is:

1. we can't just drop support for IE<11 without prod admins being ok with it.
2. we don't know what happens if mutations are not supported (are the videos still watchable? do the reference errors crash all the js code?)
3. even if 1 and 2 are resolved, the code is presented as "dirty" and doesn't feel right: why are we observing attribute changes if what we need is to apply `videojs` on new video tags? Why are `MutationRecord.type === "childList"` and `MutationRecord.addedNodes` not used instead?

ngodfraind:
1. Idk.
2. No idea. Try/catch could be used
3. Not sure it'll work. There is not a lot of doc about Mutations and I did it by trial and errors... I'll give it a try.

ngodfraind:
As you can see, using @stefk method doesn't really change a thing.

ngodfraind:
don't merge master please now it diverged

The last merge already screwed some things up.

stefk:
I don't know what we're supposed to see... Does the attribute added by videojs trigger the childList/addedNodes record? If not, the whole "infinite loop" detection stuff can be removed.

ngodfraind:
The infinite loop has to stay. I tried without and my browser crashed.

ngodfraind:
The video is a child node of something I can find in .addedNodes wich is why I run the selector on everything. It's not detected otherwise.

ngodfraind:
I'll open a new PR without the merge commits

PR #566, opened 2016-06-03
[CoreBundle] Fixing some profile url.

### LINKED ISSUE #519: [CoreBundle] Remove claro_profile_view
This route should no longer exist and be replace by claro_profile

If you do a search, it's used in notifications & badge with a userId parameter wich doesn't exist. It already wasn't working in the CoreBundle repo.

ngodfraind:
Fixes #519 

Some profile urls were wrong. Now here is the behaviour:

- if the public url is known, use it
- otherwise, use the user id to do the redirection

This fixes the notifications urls wich were just wrong (referencing a userId parameter never existed).

### DIFF
--- a/main/core/Controller/ProfileController.php
+++ b/main/core/Controller/ProfileController.php
@@ -138,14 +138,13 @@ class ProfileController extends Controller
 
     /**
      * @EXT\Route(
-     *     "/",
+     *     "/show/{user}",
      *      name="claro_profile_view"
      * )
      * @SEC\Secure(roles="ROLE_USER")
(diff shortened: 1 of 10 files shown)


stefk:
LGTM but that's the kind of thing that would need an UI approval too (screenshots etc.)

stefk:
It looks good, but shouldn't that go in the 7.x first?

ngodfraind:
It could go in the 7.x branch aswell.

stefk:
@ngodfraind, can you open a PR for 7.x instead ? We'll make a 7.x->master merge later, this will be way more simpler.

PR #585, opened 2016-06-07
[CoreBundle] Home tool refactoring

kitan1982:
- Refactors desktop Home tool
- Refactors workspace Home tool
- Refactors admin Home management

TODO :
- Tests
- Fix bug with "core_resource_logger" widget


stefk:
Controllers are way too big (+1000 lines). Using serialization tools provided by the framework would probably reduce their size by a large margin.

Also, can you describe what this PR does (features, api changes, etc.) ? It doesn't seem to be just a code refactoring.

kitan1982:
What it does :
- Refactoring of desktop & workspace Home tool with angular
- Refactoring of admin Home management tool with angular
- Changes Home tab look (Portfolio tab style)
- Allows to set a color for a Home tab
- Allows to set a color for text title in a widget

Almost all methods related to Home tool display & management have been modified to return serialized datas to work with Home tool angular version

stefk:
Ok, thanks. By serialization tools I meant things like JMS\Serializer or simply JsonSerializable, which allow you to define how the data should be serialized directly on the corresponding entities.

kitan1982:
Yes I had understood.
I will do it later.
I didn't do it on the first hand because I wanted to personalize the format and labels and it was faster this way. :D

kitan1982:
The commit message is incorrect.
I have refactored the API methods with JMS serializer

PR #590, opened 2016-06-07
[CoreBundle] Upgrade try to restore not properly uninstalled plugins.

ngodfraind:
Read code comments.
I still need to test it.

### DIFF
--- a/main/core/Controller/ResourceController.php
+++ b/main/core/Controller/ResourceController.php
@@ -572,7 +572,9 @@ class ResourceController
         $response->headers->set('Content-Transfer-Encoding', 'octet-stream');
         $response->headers->set('Content-Type', 'application/force-download');
         $response->headers->set('Content-Disposition', 'attachment; filename='.urlencode($fileName));
-        $response->headers->set('Content-Type', $mimeType);
+        if ($mimeType !== null) {
+            $response->headers->set('Content-Type', $mimeType);
+        }
         $response->headers->set('Connection', 'close');
         $response->send();
 
--- a/main/core/Library/Installation/OperationExecutor.php
+++ b/main/core/Library/Installation/OperationExecutor.php
@@ -138,20 +138,17 @@ class OperationExecutor
                     $previousPackage = $this->findPreviousPackage($bundle);
 
                     if ($previousPackage && $foundBundle) {
-                    var_dump('proper update ' . $bundle);
                         $operations[$bundle] = new Operation(Operation::UPDATE, $currentPackage, $bundle);
                         $operations[$bundle]->setFromVersion($previousPackage->getVersion());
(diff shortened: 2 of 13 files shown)


ngodfraind:
This requires the v6 package detection to be updated the same way.

ngodfraind:
@claroline/maintainers 

A much loser upgrade could be usefull imo.
If a plugin update crashes, the whole installation stops. Same for the migrations.
I think it could be great if the error was caught, then the update would still do as much as it could with a report of what didn't work for fixing afterwards.

I'd like something like this:

`php app/console claroline:update --reporting-file='/path/to/report/file' --ignore-errors`
If there were an error, the previous-installed.json file isn't removed.

Everything is done as best as possible, and errors could be found in the reporting file for manual fixes.

Errors would be things like:
- Bundle/aaa not installed because it was already installed
- Bundle/bbb not updated because it's not installed
- Bundle/ccc Error thrown at...
- Bundle/ddd Migrations couldn't be upgrade (database exists).
...

This would make upgrade easier, especially when it's a long one (I do 2x platforms upgrades with a script and can't really follow every logs with that many platforms. That's why doing everything as good as it could with the report file would be great).

ngodfraind:
Maybe make a new PR for my previous post.

stefk:
Making the upgrade process fault-tolerant is a very good idea. Here are some considerations though:

- That development isn't a bugfix at all, it might require some time and testing to get it right , and thus it should be done on the master branch.
- Being able to recover from errors shouldn't prevent us to diagnose and fix bugs or design errors in the update components themselves (which are responsible for most crashes during updates).
- We must ensure that pursuing the update won't put the platform in a completety wrong state, e.g. by disabling affected bundles, checking no information is lost in previous-installed.json, etc. If we fail to do that, a complete rollback of the upgrade is probably preferable.

I'm not in favour of the proposed `--reporting-file` and `--ignore-errors` arguments:

- reporting file: getting a report file is as easy as piping the output of the command into a file, or even into another program, syslog, whatsoever. If the output is too large -- but I don't think it is -- an option could be to use verbosity levels provided by the Console component (-vvv for full output, -vv for relatively verbose output, etc., defaulting to a minimal, informative log).
- ignore-errors: that kind of behaviour shouldn't be opt-in IMO. It should replace the previous behaviour and become the normal way of handling platform updates, providing the command:
    - ends with a big red error message informing the user that something went wrong and that if he cannot fix it,       he should rollback the update
    - exits with a status code != 0

ngodfraind:
I'm ok with that. This can still be merged. It only make the updater smarter in case something was corrupted (but it's not perfect yet).

stefk:
LGTM

stefk:
LGTM

PR #595, opened 2016-06-08
[CoreBundle] Implemented portal and search functionnalities for platform

ptsavdar:


main/core/Controller/Administration/PortalConfigurationController.php
```diff
@@ -0,0 +1,70 @@
...
+            'portalResources' => $portalManager->getPortalEnabledResourceTypes(),
+        );
+        $form = $this->createForm('portal_configuration_form', $data);
+        if ($request->isMethod('POST')) {
+            $form->handleRequest($request);
+            if ($form->isValid()) {
+                try {
+                    $portalResources = $form['portalResources']->getData();
+                    $portalManager->setPortalEnabledResourceTypes($portalResources);
+                } catch (UnwritableException $e) {
```
stefk: There's no `use` statement for that exception class.

main/core/Controller/Administration/PortalConfigurationController.php
```diff
@@ -0,0 +1,70 @@
...
+                    $form->addError(
+                        new FormError(
```
stefk: Same here.

main/core/Controller/Administration/PortalConfigurationController.php
```diff
@@ -0,0 +1,70 @@
...
+                            $this->translator->trans(
```
stefk: Not sure there's a `translator` property either. Actually the whole `catch` looks like a copy/paste.

main/core/Doctrine/Query/IfElse.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace Claroline\CoreBundle\Doctrine\Query;
+
+use Doctrine\ORM\Query\AST\Functions\FunctionNode;
+use Doctrine\ORM\Query\Lexer;
+use Doctrine\ORM\Query\Parser;
+use Doctrine\ORM\Query\SqlWalker;
+
+class IfElse extends FunctionNode
```
stefk: Is this a DQL extension? Can you explain how it works?

main/core/Resources/modules/common/register.js
```diff
@@ -0,0 +1,185 @@
+/**
+ * Created by panos on 4/5/16.
+ */
+import angular from 'angular/index'
+
+let _app = new WeakMap()
+export default class Register {
```
stefk: What's the benefit of using this register class instead of simply chaining usual methods on the `angular` object?

main/core/Resources/modules/portal/index.controller.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * Created by panos on 5/30/16.
+ */
+let _portalService = new WeakMap()
```
stefk: Why do we need weak maps for external dependencies?

ptsavdar:
@stefk First of all, I'm sorry about the missing use statements. The code was a copy/paste indeed but I thought I had imported the necessary classes. So now I fixed that and re-committed.

> Is this a DQL extension? Can you explain how it works?

Yes it is a dql extension that implements the IF(condition, onTrue, onFalse) function on MySql. I'm using it in PortalRepository to calculate a somewhat more "sofisticated" score for search results with a multiple terms query.

> What's the benefit of using this register class instead of simply chaining usual methods on the angular object?

The benefit from using this register is that it allows es6 classes to be passed as factories and directives easier than using the default angular object. Since using the default angular object you need to instantiate a factory and pass it as a parameter and for the directive, it will not work that easily.

> Why do we need weak maps for external dependencies?

WeakMaps are not a necessity. It's just my way of declaring private variables in es6 classes.

stefk:
Thanks for the explanations. LGTM

ptsavdar:
@claroline/commiters I've also added some common filters in Angular (Resources/modules/common/claro-filters.js). You can find the following filters:
- trans filter (translates the a text given it's domain[optional, default='platform'] and variables[optional, default={}]. e.g. ```{{ 'text_to_translate'|trans:'domain':{'var1': var1, 'var2': var2} }}```)
- trustAsHtml filter (to obtain values that will be accepted by SCE / privileged contexts)
- path filter (generates a path using the route name and params)
- asset (generates the asset absolute url given it's relative url)

ngodfraind:
These filters already all exists. 
I agree we should put them all at the same place. It's easier that way.

We can depreciate the others. Maybe make an PR for that an warn people on the mailing list once it's merged.

Elorfin:
For the filters, that's the most common one so it's cool to all use the same.
About the `trans` filter it's already available through the `angular-ui-translation` bower dependency. This library also provides the `transChoice` filter.

**Question :** do we add the the `transChoice` filter into the core and remove the library or remove the `trans` filter from the core and keep the library ?

IMHO, if we start to have common code in the core for Angular, it would be simplier to add the trans filters into it too.

ptsavdar:
@ngodfraind Yeap, I know that these filters exist in different places and different bundles. I agree with you that we need to warn people on the mailling list. I also think that not everything is included. I'm thinking of adding the config for tinymce. As I already said feel free to extend these functionnalities. I just tried to make a first step.

Elorfin:
Yes. It's basically the same as yours (in ES5) but it also provides pluralization through `transChoice` filter.
I've wrote it for PathBundle.

Also about the filters, we need to break them into sub-modules for cleaner dependencies (if you only want the `trans`, don't also grab the `trustAsHtml`).

ptsavdar:
@Elorfin Just checked the angular-ui-translation. I think we can use the filters from your bundle. They look more elaborate, as you perform more tests. Should we put them all here or remove mine from CoreBundle?

ptsavdar:
@Elorfin Yeap, I agree about breaking the filters into sub-modules. I thought of that too but since it was only 4 for the moment I just let them be in one module.

Elorfin:
yes but if we start using them as is, we will need to change our dependencies later ;).

For the trans, the problem is that your params are not in the same order than the library.

- `'string'|trans:{}:'domain'` in librarie
- `'string'|trans:'domain':{}` in your filter

If you want to use the library, you will need to change it in your code...

ptsavdar:
No problem changing my code. I'm removing the ```trans``` filter from my file. About breaking the filters in smaller modules, is it ok if I break them into 2: claro-sce-filters (fore trustAsHtml or trustAsResource etc.) and claro-path-filters (for path and asset) and put them all in a filters folder under common folder?

Elorfin:
The `asset` filter seems to already be available here : https://github.com/claroline/Distribution/blob/master/main/core/Resources/modules/asset/module.js.

The `trustAsHtml` is here : https://github.com/claroline/Distribution/blob/master/main/core/Resources/modules/html-truster/module.js.

So these two ones are not needed.

For the `path` one, I would create a `router` module using the same way `asset` and `trustAsHtml` are declared. Because using a `common` module will still require that you get all the filters when you add it as a dependency. Breaking filters into multiple files is important for readability but it's not sufficient, need to break them into separate modules.

ptsavdar:
@Elorfin I just saw that the ```path``` filter already exists here: :stuck_out_tongue: 
 https://github.com/claroline/Distribution/blob/master/main/core/Resources/modules/fos-js-router/module.js

So I'm guessing I'd have to delete my module completely. The good thing is that now I know where to find the for my next uses! Thanks @Elorfin , @ngodfraind and @stefk !

Elorfin:
I've learned it too ;).

ptsavdar:
@claroline/commiters To resume: Forget my previous comment. The filters exist already, some in the coreBundle (cudos and thanks to **FormaLibre** team) and the translation filter in ```angular-ui-translation``` package (cudos and thanks to **Innova** team). The filters are the following:

- trans inside angular-ui-translation package in webpacks (usage: ```{{ 'text_to_translate'|trans:{'var1': var1, 'var2': var2}:'domain' }}```)
- path inside coreBundle under modules/fos-js-router
- asset inside coreBundle under modules/asset
- trust_html inside coreBundle under modules/html-truster

Elorfin:
LGTM.

@ptsavdar : just to be sure, this will not be available in the 7.x branch, is that ok for you ?

PR #597, opened 2016-06-08
Allow a non file resource to be downloaded according to it's mime type

### LINKED ISSUE #576: [Plugin][VideoPlayer] Download audio resource
When attempting to download an audio file (one recorded with AudioRecorder plugin) the file has no extension...

pitrackster:
Can download a 'non file' resource that is... a file...


stefk:
LGTM.

@pitrackster, can you make the PR title more explicit (that title will become the commit message when we'll squash).

LaurentGruber:
LGTM

PR #610, opened 2016-06-09
[Webpack] Optimize prod usage and enforce Resources/modules convention

stefk:
- Dependencies that are included in at least three chunks/modules are automatically moved to a *commons.js* file. That file will contain frequently required libs, avoiding the need to bundle them again and again, and allowing proper caching on client side. It must be included before every `hotAsset` call.
- Compiled output is minified in prod. [needs a separate PR on claroline/Claroline]
- The Resources/modules directory is where ES6 code is expected to live, thus:
   - imports of modules across different bundles of the distribution are shorthened: `import foo from '#/main/core/Resources/modules/bar'` becomes `import foo from '#/main/core/bar'`
  - the `modules/` prefix must be omitted in assets.json files


main/core/Resources/views/Administration/Facet/index.html.twig
```diff
@@ -30,6 +30,5 @@
 
 {% block javascripts %}
     {{ parent() }}
-
-    <script type="text/javascript" src="{{ asset('bundles/clarolinecore/js/administration/facet.js') }}"></script>
+    <script src="{{ asset('bundles/clarolinecore/js/administration/facet.js') }}"></script>
```
Elorfin: You've added : `<script src="{{ hotAsset('dist/commons.js') }}"></script>` everywhere but not here. Is that normal ?

stefk: That one is a simple `asset`, not `hotAsset`, so it isn't needed.

Elorfin:
LGTM

ngodfraind:
LGTM

ngodfraind:
7.x branch ? it's may be annoying otherwise because this is kind of an important change.

stefk:
I don't know. IMO it shouldn't go on the 7.x as it's not a bugfix per se. @LaurentGruber?

LaurentGruber:
@stefk @ngodfraind si ça ne change rien pour l'utilisateur et que ça ne représente pas un gros risque d'introduction de nouveau bug pour moi ça peut aller dans la 7.x sinon je pense que c'est mieux dans la 16.09

ngodfraind:
C'est pas un bugfix, mais ça ne change rien pour l'utilisateur (et a priori ne causera pas de bugs).

C'est simplement pour nous simplifier la vie et ne pas avoir à rechanger les chemin '#/' dans le futur si on merge la 7.x.

stefk:
> a priori ne causera pas de bugs

A priori non, mais pourquoi tenter ? La divergence avec la 7.x est inévitable, autant s'y faire tout de suite.

PR #612, opened 2016-06-09
[ScormBundle] Full screen mode.

ngodfraind:
The allowfullscreen attribute somehow fixed some fullscreen issue with firefox & IE11 for me.

https://developer.mozilla.org/en/docs/Web/HTML/Element/iframe


ngodfraind:
For the history:

I never tought it would ever happen but the only correct behaviour was from IE11 before.
Firefox was just buggy and it's kind of a security hole for chrome.

Elorfin:
LGTM

stefk:
LGTM

PR #620, opened 2016-06-10
[CoreBundle] Add confirm modal management

Elorfin:
Maybe not needed in 7.x.

This PR contains a new directive to simplify the usage of AngularBootstrap modal.
For now only confirm modals are managed based on ngBootbox way to do it (directives that can be appended to DOM elements).

So once this PR is merged, there will be no more need of ngBootbox (which requires jquery and bootstrap js) which is only used in Exercises.

Implementation détails : 
- I have kept the controller inside the directive definition as the directive and the controller do practically nothing.
- The design and features are very limited for now (I only do what bootbox did) and will be updated when adding alert modal and other modal types.

See README for more information about usage.

### DIFF
--- a/main/core/Resources/modules/modal/Controller/ConfirmModalController.js
+++ b/main/core/Resources/modules/modal/Controller/ConfirmModalController.js
@@ -1,0 +1,14 @@
+export default class ConfirmModalController {
+    constructor($uibModalInstance, content) {
+        this.$uibModalInstance = $uibModalInstance
+        this.content = content
+    }
+
+    cancel() {
+        this.$uibModalInstance.dismiss('cancel');
+    }
+
+    confirm() {
+        this.$uibModalInstance.close();
+    }
+}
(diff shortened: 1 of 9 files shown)


stefk:
It looks nice. I think it should go on the master branch though. Other minor concern: isn't `ui.*` a kind of namespace for angular-boostrap? Shouldn't we use our own namespace for components we develop?
Something like `claro.*` or `cc.*`. ping @claroline/commiters

Elorfin:
I'm not sure they are the only one who use `ui.*`.

I have followed : https://github.com/claroline/Distribution/blob/master/main/core/Resources/modules/asset/module.js which also use the ui prefix.

+1 for having our own namespace.
This feature is really generic, it may not be needed to lock it into a `claro` namespace. I would keep the `claro` namespace for our shared features which are specific to our platform (like resource management).

Elorfin:
Closed in favor of #624.

PR #623, opened 2016-06-10
[Tinymce]  plugins loaded with webpack

ngodfraind:
The tinymce loader isn't always reliable.

```tinymce.PluginManager.load('mention', home.asset + 'packages/claroline-tinymce-mention/plugin.min.js');``` 

will crash it if it can't find the plugin.

No more problem with webpack.

@ptsavdar 
This fixes codemirror but:
- How can I test the mention plugin ?
- Can I remove the accordion plugin ?
- The math editor plugin doesn't work properly for me.

### CODE REVIEW

**Comment by LaurentGruber** [comment]
LGTM

File: main/core/Resources/public/js/tinymce/plugins/codemirror.js [hunk 1]
```diff
@@ -0,0 +1,45 @@
...
+	});
+
+	// Add a menu item to the tools menu
+	editor.addMenuItem('code', {
+		icon: 'code',
+		text: 'Source code',
+		context: 'tools',
+		onclick: showSourceEditor
+	});
+});
```
stefk: Use soft tabs instead.

ngodfraind:
- update the mention plugin version
- checkstyle

Done.

main/core/Resources/public/js/tinymce/tinymce.js
```diff
@@ -1,353 +1,326 @@
-/*
- * This file is part of the Claroline Connect package.
+import 'claroline-tinymce-mention/plugin.min'
```
stefk: Sorry, I didn't see these imports in the first review. With them, there's no point leaving this file in the public directory, because browsers won't be able to parse it. It should be moved to the *modules* directory.

ptsavdar:
@ngodfraind I think you can delete the accordion plugin. About the mention plugin you can test it in an editor inside a resource (e.g. Wiki) by typing ```@nameofuser``` . The math editor seems to work fine in my local environment. Can you give me some more details about the problem you're having?

ptsavdar:
LGTM

ngodfraind:
- Tinymce moved to the modules directory.
- Accordion plugin removed

Everything is working as intended.

stefk:
LGTM

PR #624, opened 2016-06-10
[CoreBundle] Add confirm modal management

Elorfin:
Original PR was #620 but it's not needed for the 7.x.

This PR contains a new directive to simplify the usage of AngularBootstrap modal.
For now only confirm modals are managed based on ngBootbox way to do it (directives that can be appended to DOM elements).

So once this PR is merged, there will be no more need of ngBootbox (which requires jquery and bootstrap js) which is only used in Exercises.

Implementation détails : 
- I have kept the controller inside the directive definition as the directive and the controller do practically nothing.
- The design and features are very limited for now (I only do what bootbox did) and will be updated when adding alert modal and other modal types.

See README for more information about usage.


stefk:
LGTM

Elorfin:
LGTM

PR #627, opened 2016-06-10
[CoreBundle] Display profile links in administration and contact.

ngodfraind:
Correct branch.

Still need to work on the privacy settings


ngodfraind:
Done.

main/core/Controller/API/User/UserController.php
```diff
@@ -255,6 +255,52 @@ public function getUserAction(User $user)
...
+     * @Get("/user/{user}/public", name="get_public_user", options={ "method_prefix" = false })
+     */
+    public function getPublicUserAction(User $user)
+    {
+        $settingsProfile = $this->profilePropertyManager->getAccessesForCurrentUser();
+        $publicUser = [];
+
+        foreach ($settingsProfile as $property => $isEditable) {
+            if ($isEditable || $user === $this->container->get('security.token_storage')->getToken()->getUser()) {
+                switch ($property) {
```
Elorfin: It seems you can put the `$user` check outside of the loop.

Elorfin:
LGTM

main/core/Tests/API/User/UserControllerTest.php
```diff
@@ -496,6 +496,39 @@ public function testCsvImportFacetsAction()
...
+        $data = json_decode($this->client->getResponse()->getContent(), true);
+        $this->assertEquals(['username' => 'admin'], $data);
+
+        $this->logIn($admin);
+        $this->client->request('GET', "/api/user/{$user->getId()}/public");
+        $data = json_decode($this->client->getResponse()->getContent(), true);
+        $this->assertEquals(
+            ['email' => 'user@mail.com', 'firstName' => 'user', 'lastName' => 'user', 'username' => 'user'],
+            $data
+        );
```
stefk: Would be better/more readable to have distinct tests for these scenarios:
- an admin can see any profile
- a user can see its own profile
- a user can see fields of other profiles fields only if granted explicitely

ngodfraind: What do you think about more comments ?
Currently, it's pretty easy to see wich method is tested because every test is done the same way:

```
testMyFunction() {
}
```

And then it handles every use cases.
Also (but it's minor) I only need to create the user/admin once.

Is it better to multiplicate tests or do better documentation ?

ptsavdar:
LGTM

main/core/Tests/API/User/UserControllerTest.php
stefk: Ideally tests should be short and self-documenting: an explicit name, a quick setup and a couple of assertions.

stefk:
LGTM

Issue #463, opened 2016-05-30
Matching question


marion42:
For matching to bind questions, when we associated multiple proposal with multiple label and we don't put the correct answer, the notation isn't correct.

PR #637, opened 2016-06-13
Fix claroline/Distribution#463 Fix scoring on match questions

eldoniel:
This commit fixes the following issue: 
https://github.com/claroline/Distribution/issues/463

Now the score set for a label is counted for each right connection.


Elorfin:
LGTM

stefk:
LGTM

PR #638, opened 2016-06-13
[Installation] Catch the error if claroline isn't installed yet while…

ngodfraind:
… doing the plugin check.


Elorfin:
LGTM

purplefish32:
LGTM

PR #657, opened 2016-06-15
Exo issues fixes

eldoniel:
This PR fixes two minor issues.

On graphic questions, since the pointers are no longer shown on page load, but appear on click, it seemed important to say it in a descriptive sentence.

On match question, the correction directive wasn't loading; the problem is fixed. The correction used to show a column "your answer", which is useless because we always show it with the question directive, containing these answers.


Elorfin:
LGTM

stefk:
Needs a rebase, otherwise LGTM.

Elorfin:
LGTM

PR #664, opened 2016-06-16
Exo qti import 7.x

david42:
To validate each qti file in the import


plugin/exo/Services/classes/QTI/OpenImport.php
```diff
@@ -70,4 +75,59 @@ protected function getPrompt()
...
+                break;
+            case 'short':
+                $validated = $this->shortQtiValidate();
+                break;
+            case 'oneWord':
+                $validated = $this->oneWordValidate();
+                break;
+        }
+
+        return $validated;
```
stefk: A `default` case is missing in the switch, or `$validated` should have an initial value.

david42:
@stefk Yes, thanks.
I added a default value

stefk:
LGTM

LaurentGruber:
LGTM

PR #697, opened 2016-06-20
[CoreBundle] Honeypot form basic implementation.

ngodfraind:
This is a (very) basic implementation of a honeypot I'd like to try (and combine it with the captcha to reduce the amount of bots going through).

It can be improved later. I plan to use it on claroline.net and see how it goes.

My prefered method for catching bot works with javascript (it's simply adding a parameter in the querystring when the user clicks on a submit button... this parameter is then checked and if it doesn't exists, we assume it's a bot) but it's not very generic and I don't think it'll be usable on a large scale.

it's related to #388 but honeypots are less of an accessibility problem.


main/core/Form/Field/HoneypotType.php
```diff
@@ -0,0 +1,90 @@
...
+            'required' => false,
+            'mapped' => false,
+            'data' => '',
+            'label' => ' ',
+            'attr' => array(
+                'autocomplete' => 'off',
+                'tabindex' => -1,
+                // Fake `display:none` css behaviour to hide input
+                // as some bots may also check inputs visibility
+                'style' => 'position: fixed; left: -100%; top: -100%;',
```
stefk: Accessiblity issue. Screen readers users should be warned somehow that the field must be left empt, eg by using an appropriate labe/aria-label (probably without giving a too obvious hint to bots).

stefk:
What's the related issue? I can't be #366.

ngodfraind:
it's #388 sorry

ptsavdar:
Besides accessibility issue mentioned by @stefk I'd say LGTM

LaurentGruber:
LGTM

LaurentGruber:
LGTM

PR #702, opened 2016-06-21
[CoreBundle] Removing update buggy code

ngodfraind:
This didn't work well ($bundle not initialized; I don't get why it worked at some point). I guess we'll wait to work on the much tolerant upgrade to fix this properly. 

It would be nice to be able to find if the plugin is already in the database but that's not the correct way to do it.


Elorfin:
LGTM

ngodfraind:
LGTM

Elorfin:
LGTM

PR #709, opened 2016-06-22
[PathBundle] Summary enhancements + progression bar

ptsavdar:


ptsavdar:
Adds a progress bar on the top of the summary frame. Summary can now get pinned/unpinned in/from the content.

Elorfin:
Why do you make the calculations on server side. 
All the data (count steps, count progression) are directly available in Angular, so why do not directly use them ?

ptsavdar:
You mean the total progression? Yes the total steps and the total progression are already there when user loads the page but, let's suppose that the user chooses a path and marks it as "to review" or "to do", then we need to change the total progression but then again the user decides to "rechange" the status to "to review" or "to do" or "done" or "seen" then I'd have to either increment or leave the same. And in order to avoid keeping track of every step in js I preferred to ask the server for this information. To avoid any miscalculation or having js go through all steps to recalculate total progress. (it costs less to server than the client IMO). What do you think? If you want we can change that in the future, it's fine by me. But right now I don't know if I would have the time to make this change.

Elorfin:
I'm OK to let it in the server side for now.
But I don't think we need to keep track of every step for the progression, just updating the value when the progression is updated.

Anyway Path will need some refactoring to integrate webpack and use core's libraries, so we can see this at the same time.

ptsavdar:
@Elorfin Ok I'll tranfer logic to Angular when I'll have the time.

stefk:
Closed in favour of #709

PR #715, opened 2016-06-22
Merge 7.x into master

stefk:
No big conflicts, no migration clashes apparently.

@Elorfin, @eldoniel, I'm unsure about a change that happened in the 7.x. Correction directives were renamed and this caused a conflict for match questions: `correctionMatch` vs `matchCorrection`. I don't know which one must be chosen.


eldoniel:
I did the change. correctionMatch was used, but the right name of the directive being matchCorrection, the directive wasn't called (causing the correction for the match question to be systematicly empty).
So I guess the right one should be matchCorrection.

Elorfin:
@stefk : I have commited the correct one (as said by @eldoniel, it is matchCorrection).

stefk:
Thanks. Forgot to mention this is a no-squash PR.

Elorfin:
It seems there is also a bug in Match Questions in exo (can not reproduce it on 7.x branch). But for now, I don't see why it doesn't work. It sould !

PR #717, opened 2016-06-23
[CoreBundle] Avoid creating a tmp directory with the template filename

stefk:
Currently, when creating a user, the workspace manager tries to extract the personal.zip archive into a directory called personal.zip...


Elorfin:
LGTM

purplefish32:
LGTM

PR #718, opened 2016-06-23
[ExoBundle] Get the step definition when added to Exercise

Elorfin:
When adding a Step to an Exercise, the whole definition of the created step is not set in Angular. So if you try to edit the Step meta directly after angular crash.

Not a big bug because the data are correctly saved.


stefk:
LGTM

ptsavdar:
LGTM

Elorfin:
LGTM

PR #723, opened 2016-06-23
[FlashCardBundle] Fixing flashcard creation.

ngodfraind:
- Uses the core select form for the buggy select.
- Select improved (it has a choice_value option and will select the by default object).
- Redirect at the note creation.

### DIFF
--- a/plugin/flashcard/Resources/modules/flashcard/flashcard.js
+++ b/plugin/flashcard/Resources/modules/flashcard/flashcard.js
@@ -1,6 +1,6 @@
 /*
  * This file is part of the Claroline Connect package.
- * 
+ *
  * (c) Claroline Consortium <consortium@claroline.net>
  *
  * For the full copyright and license information, please view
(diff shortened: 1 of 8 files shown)


plugin/flashcard/Resources/modules/flashcard/flashcard.js
```diff
@@ -12,6 +12,7 @@ import angular from 'angular/index'
 import {} from 'angular-bootstrap'
 import {} from 'angular-route'
 import ClarolineAPI from '../../../../../main/core/Resources/modules/services/module'
```
Elorfin: You can simplify the path.

ngodfraind: I know but I didn't plan to refactor the whole bundle -,-.

ngodfraind: There are also some other minor issues (a lot of unused $http, for (...;;;;) instead of forEach). It should be done in a future pr but not yet.

Elorfin: I understand. I have seen it just because it's over a line you've added ;).

Elorfin:
LGTM

ngodfraind:
LGTM

Issue #593, opened 2016-06-08
[PathBundle] Restart path on the last visited step


Elorfin:
When a User open a Path he has already started, we need to send him directly on the last he has seen.

@eddymarques : FYI.

PR #727, opened 2016-06-24
[PathBundle] Restart path on the last seen step

Elorfin:
In order to fix the #593.

In the player if no step is specified in the URL (default behavior when you open a Path), the User is sent to the last Step he has seen or the last step he has set a manual progression (done / to do / to redo).

Implementation remark : the `wizard.app.js` starts to become quite ugly. I will need to split it to a player and editor entry points as the behavior of the router is no longer the same in the 2 apps. Will require some works and tests so it will be done later.

ping @eddymarques @ptsavdar


stefk:
LGTM

ptsavdar:
LGTM

Issue #693, opened 2016-06-20
Improve automatic checkstyle on PRs


stefk:
As a significant amount of time is still lost on minor style details when reviewing PRs, I'd like to propose enforcements/enhancements to our existing setup:

- [x] Force PHP shortand array notation (there's a `short_array_syntax` fixer in php-cs-fixer's contribs for that)
- [x] Force PHP strict comparisons (~~see `strict` contrib in php-cs-fixer~~)
- [x] Force ordering use statements (idem, see `ordering_use_statements`)
- [ ] Find a way to detect wrong indents (current standard isn't severe enough)
- [x] Add a checkstyle for js files, at least for the es6 files living in *Resources/modules* (that check could be based on a full ruleset like [standard](http://standardjs.com/) or begin as a minimal [eslint](http://eslint.org/) config that would be improved along the way)

 **EDIT**: strict comparisons shoudn't be handled by phpcs-fixer (see below)

CPASimUSante:
Is there a reason other than gaining 5 characters to use [] instead of array() ?

purplefish32:
@CPASimUSante : consistancy

stefk:
No, it's just shorter and easier to spot. That notation can also be found in many other languages.

purplefish32:
LGTM

arnaudbey:
:+1:

ngodfraind:
LGTM

2016-06-20 13:09 GMT+02:00 Arnaud Bey <notifications@github.com>:

> 👍
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/claroline/Distribution/issues/693#issuecomment-227113878>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ABVStrT9Dh8Xa2vpzhl0FR59WcUFyiqYks5qNnTOgaJpZM4I5inx>
> .
>



-- 
Godfraind Nicolas
5, rue de la Carrière
6592 Monceau-Imbrechies
GSM: 0472 94 80 99

Elorfin:
> Force PHP strict comparisons (see strict contrib in php-cs-fixer)

This can introduce bugs for code which use loose comparison to handle different types (generally used for inconsistent booleans).

IMHO, this should be spotted but not automatically replaced.

stefk:
[...]

PR #728, opened 2016-06-24
Add ESLint check on ES6 files

stefk:
See #693. 

- the check will be applied to js files within *Resources/modules/** directories
- standard used is described in https://github.com/claroline/Distribution/issues/693#issuecomment-227124134 (I've added single quotes preference and unix-style line endings as per ESLint suggestion)

Note that this PR just adds the check, it doesn't fix any actual errors. Those will have to be fixed on each modified files prior to a PR submission. 

There're currently 8000+ infractions to the standard. Most of them are indent and semi-colons issues that can be fixed automatically by ESLint using the `--fix` option. The remaining ones are mainly:

- unknown globals (ESLint requires globals to be specified explicitely to prevent leakage)
- missing import statements
- unused variables
- variables defined more than once
- console statements

cc @claroline/commiters

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,16 @@
+{
+  "env": {
+    "browser": true,
+    "es6": true
+  },
+  "extends": "eslint:recommended",
+  "parserOptions": {
+    "sourceType": "module"
+  },
+  "rules": {
+    "indent": [2, 2],
+    "linebreak-style": [2, "unix"],
+    "quotes": [2, "single"],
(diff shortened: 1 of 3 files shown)


stefk:
Failure output example: https://travis-ci.org/claroline/Distribution/builds/139998021#L1753

stefk:
Good to merge IMO.

Elorfin:
LGTM

stefk:
LGTM

PR #731, opened 2016-06-24
[PathBundle] Fixed view for anonymous and tranfered progression logic to angular

ptsavdar:


ptsavdar:
LGTM (I know it's snicky but I'm in a hurry) :stuck_out_tongue:

plugin/path/Entity/UserProgression.php
```diff
@@ -198,7 +198,7 @@ public function jsonSerialize()
     {
         return array(
             'id' => $this->id,
-            'userId' => $this->user->getId(),
+            'userId' => ($this->user instanceof User)?$this->user->getId():0,
```
Elorfin: I think this is not needed as the progression is not saved for anonymous.

ptsavdar: Yes, I know but in order to avoid the 500 response I decided to send one as response. Helps me calculate the totalprogression.

nicolasjacquemet:
LGTM

eddymarques:
LGTM

Elorfin:
LGTM

PR #736, opened 2016-06-27
[PathBundle] improve iFrame resize

dtisch:
Integration of preliminary work done by Elorfin and temporary hack for being able to react to iFrame inner window height change initiated by an angular request.


stefk:
Commented code and log statements must be removed.

stefk:
@dtisch, are you done here?

dtisch:
@stefk, yes I'm done. Did I forget to do something or warn someone?

main/core/Resources/views/Workspace/layout.html.twig
```diff
@@ -90,6 +90,38 @@
...
+                parent.postMessage(message, "*");
+            }
+        };
+
+        $(document).ready(function () {
+            postHeight();
+
+            /* Hack for constantly polling iframe inner height and react when angular router loads another page */
+            var getIframeWindowHeight = function() {
+                return $("iframe#embeddedActivity").contents().find('body').first().height();
```
Elorfin: Can't you just rely to `$(document).height()` instead of using the iFrame as a selector ? So you didn't need to know how the resource is displayed in the page and it will work in any case (not only in path).

stefk:
No, it's just the words "temporary" which made me unsure :) I thought maybe something else was coming.

LGTM

main/core/Resources/views/Workspace/layout.html.twig
dtisch: My ugly hack is not intended to work in any case, but only in the case when displayed in a step of the PathBundle, and only until a better solution involving an hypothetical "angular has finished loading the new page" event is found.
I'm targeting the iFrame directly for a second reason: the height of the iFrame has to be adjusted when the angular module inside the iFrame loads a page bigger or smaller than the previous one displayed in the iFrame. I suppose that if I'm relying on $(document).height(), I will not detect any height change because the Iframe, fitted with a scrollbar if needed, won't change its parent height.

LaurentGruber:
ça tiens toujours compte du paramètre min-height qu'on peut fixer dans la step ?

dtisch:
Oui, le style CSS min-height est conservé (toujours en CSS inline dans la balise `iframe`).

main/core/Resources/views/Workspace/layout.html.twig
Elorfin: > I suppose that if I'm relying on $(document).height(), I will not detect any height change because the Iframe, fitted with a scrollbar if needed, won't change its parent height.

This was the underlaying question in my comment : do you get the correct height with `$(document).height()` in an IFrame ? ;). I have never tested it and I wanted to know if it works.

I'm OK with keeping the code like that, as you have stated, it will need to be replaced by a better implementation later.

Elorfin:
@dtisch : the Laurent's question is not related to CSS. In the Editor, you can set a "Resource height" property in the Steps. As the old resize method did not work well, it is a workaround : you can set a huge height to be sure there will be no scrollbar.

@LaurentGruber : I haven't tested it but the implementation basically works the same way the older (that's just the way it get the correct height which have changed), so it sould not be a problem. 

Also, the new implementation will always get the correct height, so the min height will become obsolete.

dtisch:
@Elorfin, @LaurentGruber  : I know, the min-height is defined in the step configuration form but as it ends in the style attribute of the iframe tag, I considered it as inline CSS :-) I didn't touch this option but as you said, it will become obsolete. With the hack, it's not required to use this option but if so, it will be taken into consideration by the browser.

`<iframe id="embeddedActivity" style="width: 100%; min-height: 3000px;" data-ng-src="/app_dev.php/resource/open/icap_lesson/7" allowfullscreen="" data-resources="stepShowCtrl.step.primaryResource" data-height="stepShowCtrl.step.activityHeight" class="ng-isolate-scope" src="/app_dev.php/resource/open/icap_lesson/7"></iframe>`

Elorfin:
LGTM

PR #740, opened 2016-06-27
Update composer.json

ngodfraind:
rest-bundle 1.8.0 doesn't work with jms serializer (it crashed everything for me, probably because the jms serializer version wasn't the correct one anymore).

This is the easy fix.


LaurentGruber:
LGTM

Elorfin:
LGTM

PR #743, opened 2016-06-28
[ExoBundle] Fix question sharing and categories management

Elorfin:
This PR fixes the saring feature of the Bank which was buggy in the last version.
It also fixes the #737. 

I've just made the existing code work, but the feature seems a little weird to me : 

The categories are specific to a User. 
If you edit a question which have been shared with you it loads the list of your categories.
So the previous category doesn't exist (that's the origin of the original bug). 
In this case, the implementation replaces the original category by a default one linked to the current User.

The weird part is it overrides the original categorization made by the User who have shared the question.

As we will need to rewrite the Bank in Angular for the 16.09, I have kept it but IMHO we will need to rethink this part.

### DIFF
--- a/plugin/exo/Controller/InteractionGraphicController.php
+++ b/plugin/exo/Controller/InteractionGraphicController.php
@@ -227,9 +227,14 @@ class InteractionGraphicController extends Controller
         );
 
         $formHandler = new InteractionGraphicHandler(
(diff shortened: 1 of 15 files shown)


stefk:
LGTM

LaurentGruber:
LGTM and ok with the comment about the feature

PR #744, opened 2016-06-28
Exo question add fields

eldoniel:
This adds some fields we used in V5 at innovalangues.

The fields are set as follow:
- Functional instruction is set on the "description" field.
- Object of the question uses the ObjectQuestion entity, refering to a question and a resourceNode.
- Pedagogic instruction is set on the "specification" field.
- Complementary information is set on the "supplementary" field.

This PR doesn't update the editor in order to fill these fields, it only allows to show existing data in the player.

### DIFF
--- a/plugin/exo/Entity/ObjectQuestion.php
+++ b/plugin/exo/Entity/ObjectQuestion.php
@@ -21,7 +21,7 @@ class ObjectQuestion
 
     /**
      * @ORM\Id
-     * @ORM\ManyToOne(targetEntity="UJM\ExoBundle\Entity\Question")
+     * @ORM\ManyToOne(targetEntity="UJM\ExoBundle\Entity\Question", inversedBy="objects")
      * @ORM\JoinColumn(onDelete="CASCADE")
      */
     private $question;
@@ -31,7 +31,7 @@ class ObjectQuestion
      *
      * @ORM\Column(name="ordre", type="integer")
      */
-    private $ordre;
+    private $order;
 
     /**
      * @ORM\Column(type="text", nullable=true)
@@ -61,7 +61,7 @@ class ObjectQuestion
         return $this->exercise;
     }
 
-    public function setQuestion(Question $question)
+    public function setQuestion(Question $question = null)
     {
         $this->question = $question;
     }
@@ -74,11 +74,11 @@ class ObjectQuestion
     /**
      * Set ordre.
      *
-     * @param int $ordre
+     * @param int $order
      */
-    public function setOrdre($ordre)
+    public function setOrdre($order)
     {
-        $this->ordre = $ordre;
+        $this->order = $order;
     }
 
     /**
@@ -88,7 +88,7 @@ class ObjectQuestion
      */
     public function getOrdre()
     {
-        return $this->ordre;
+        return $this->order;
     }
 
     /**
@@ -121,5 +121,21 @@ class ObjectQuestion
     public function getType()
     {
         return $this->type;
+    }
+
+    /**
+     * @return ResourceNode
+     */
+    public function getResourceNode()
+    {
+        return $this->resourceNode;
+    }
+
+    /**
+     * @param ResourceNode $resourceNode
+     */
+    public function setResourceNode(ResourceNode $resourceNode)
+    {
+        $this->resourceNode = $resourceNode;
     }
 }
--- a/plugin/exo/Entity/ObjectQuestion.php
+++ b/plugin/exo/Entity/ObjectQuestion.php
@@ -3,6 +3,7 @@
 namespace UJM\ExoBundle\Entity;
 
 use Doctrine\ORM\Mapping as ORM;
+use Claroline\CoreBundle\Entity\Resource\ResourceNode;
 
 /**
  * UJM\ExoBundle\Entity\ObjectQuestion.
--- a/plugin/exo/Entity/ObjectQuestion.php
+++ b/plugin/exo/Entity/ObjectQuestion.php
@@ -2,8 +2,8 @@
 
 namespace UJM\ExoBundle\Entity;
 
-use Doctrine\ORM\Mapping as ORM;
 use Claroline\CoreBundle\Entity\Resource\ResourceNode;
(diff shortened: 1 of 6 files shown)


plugin/exo/Entity/Question.php
```diff
@@ -119,7 +125,7 @@ public function setType($type)
         $this->type = $type;
     }
 
-    /**
+    /**, orphanRemoval=true
```
stefk: Should be removed.

plugin/exo/Manager/QuestionManager.php
```diff
@@ -96,13 +102,32 @@ public function importQuestion(\stdClass $data)
...
+        $data->supplementary = $question->getSupplementary();
+        $data->specification = $question->getSpecification();
+        $data->objects = array_map(function ($object) use ($rm) {
+            $resourceObjectData = new \stdClass();
+            $resourceObjectData->id = (string) $object->getResourceNode()->getId();
+            $resourceObjectData->type = $object->getResourceNode()->getMimeType();
+            $resourceObjectData->name = $object->getResourceNode()->getName();
+            $resourceObjectData->path = $object->getResourceNode()->getPathForDisplay();
+            switch ($object->getResourceNode()->getMimeType()) {
+                case 'custom/hevinci_url':
```
stefk: Is this really necessary? Exo plugin shouldn't be aware of url plugin IMO.

plugin/exo/Entity/ObjectQuestion.php
```diff
@@ -122,4 +122,20 @@ public function getType()
...
+     */
+    public function getResourceNode()
+    {
+        return $this->resourceNode;
+    }
+
+    /**
+     * @param ResourceNode $resourceNode
+     */
+    public function setResourceNode(ResourceNode $resourceNode)
```
Elorfin: use statement for `ResourceNode` is missing.

@stefk : why is this not catched by automatic analysis ? Isn't it part of the new config ?

plugin/exo/Manager/QuestionManager.php
```diff
@@ -96,13 +102,32 @@ public function importQuestion(\stdClass $data)
...
         $data->title = $question->getTitle();
         $data->description = $question->getDescription();
         $data->invite = $question->getInvite();
+        $data->supplementary = $question->getSupplementary();
+        $data->specification = $question->getSpecification();
+        $data->objects = array_map(function ($object) use ($rm) {
+            $resourceObjectData = new \stdClass();
+            $resourceObjectData->id = (string) $object->getResourceNode()->getId();
+            $resourceObjectData->type = $object->getResourceNode()->getMimeType();
+            $resourceObjectData->name = $object->getResourceNode()->getName();
```
Elorfin: This is not needed and not part of the JSON Schema (see `Content` section).

plugin/exo/Manager/QuestionManager.php
```diff
@@ -96,13 +102,32 @@ public function importQuestion(\stdClass $data)
...
+            $resourceObjectData->path = $object->getResourceNode()->getPathForDisplay();
```
Elorfin: Idem

plugin/exo/Manager/QuestionManager.php
```diff
@@ -96,13 +102,32 @@ public function importQuestion(\stdClass $data)
...
+            switch ($object->getResourceNode()->getMimeType()) {
+                case 'custom/hevinci_url':
+                    $resourceObjectData->resourceUrl = $rm->getResourceFromNode($object->getResourceNode())->getUrl();
```
Elorfin: Property needs to be named `url` : http://json-quiz.github.io/json-quiz/spec/content.html.

plugin/exo/Manager/QuestionManager.php
```diff
@@ -96,13 +102,32 @@ public function importQuestion(\stdClass $data)
...
+                    break;
+                default:
+                    $resourceObjectData->resourceUrl = $rm->getResourceFromNode($object->getResourceNode())->getHashName();
```
Elorfin: The default case needs to be the real URL to access the Resource (from memories, the route claro_resource_open).

Elorfin:
You also need to manage the case of HTML resources which use an attribute `data` (instead of `url`) in exports.

plugin/exo/Resources/modules/question/Partials/show.html
```diff
@@ -45,6 +45,10 @@ <h4 class="panel-title">
             <!-- Description of the Question -->
             <div data-ng-bind="questionShowCtrl.question.description"></div>
         </div>
+        <div class="question-info" data-ng-if="questionShowCtrl.question.objects.length > 0">
+            <h5>{{ 'question_object'|trans:{}:'ujm_sequence' }}</h5>
+            <div data-ng-repeat="object in questionShowCtrl.question.objects">{{ object.resourceUrl }}</div>
```
Elorfin: For content with `url` property, this needs to be a clickable link.

plugin/exo/Entity/ObjectQuestion.php
stefk: It only detects uses of fully-qualified class names ( ≃ classes starting with `\`, without `use` statements). This isn't the case here. As far as the tool is concerned, `ResourceNode` is just a class in the `UJM\ExoBundle\Entity` namespace.

stefk: An automatic test would have caught it, tough...

plugin/exo/Resources/translations/ujm_sequence.fr.yml
```diff
@@ -92,6 +92,9 @@ points: Points
 question_hints: Indices
 question_hint_label: Indice
 question_hint_show_title: Voir l'indice
+question_object: Objet de la question
+question_specification: Specification
+question_supplementary: Supplementary information
```
LaurentGruber: ça devrait être du FR

LaurentGruber:
LGTM

purplefish32:
LGTM

PR #745, opened 2016-06-28
[CoreBundle] Conversion from form data array to User entity

### LINKED ISSUE #741: Création utilisateur lien d'inscription publique EA
La création du compte en utilisant le lien d'inscription publique d'un EA provoque une erreur.
La création du même utilisateur via l'interface administrateur ne pose pas de problème.

image

Stack Trace
in vendor/claroline/distribution/main/core/Manager/UserManager.php at line 145   -
     * @return \Claroline\CoreBundle\Entity\User
     */
    public function createUser(
        User $user,
        $sendMail = true,
        $rolesToAdd = array(),
        $model = null,
at ErrorHandler ->handleError ('4096', 'Argument 1 passed to Claroline\CoreBundle\Manager\UserManager::createUser() must be an instance of Claroline\CoreBundle\Entity\User, array given, called in /data/www/clarolineconnect/claroline/vendor/claroline/distribution/main/core/Controller/Tool/WorkspaceParametersController.php on line 334 and defined', '/data/www/clarolineconnect/claroline/vendor/claroline/distribution/main/core/Manager/UserManager.php', '145', array()) 
in vendor/claroline/distribution/main/core/Manager/UserManager.php at line 145   + 
at UserManager ->createUser (array('firstName' => 'dinora', 'lastName' => 'dinora', 'username' => 'dinora', 'plainPassword' => 'noradi', 'mail' => 'dinora@dinora.com', 'locale' => 'fr', 'accepted_terms' => true)) 
in vendor/claroline/distribution/main/core/Controller/Tool/WorkspaceParametersController.php at line 334   + 
at WorkspaceParametersController ->anonymousSubscriptionAction (object(Workspace)) 
at call_user_func_array (array(object(WorkspaceParametersController), 'anonymousSubscriptionAction'), array(object(Workspace))) 
in app/bootstrap.php.cache at line 3109   + 
at HttpKernel ->handleRaw (object(Request), '1') 
in app/bootstrap.php.cache at line 3071   + 
at HttpKernel ->handle (object(Request), '1', true) 
in app/bootstrap.php.cache at line 3222   + 
at ContainerAwareHttpKernel ->handle (object(Request), '1', true) 
in app/bootstrap.php.cache at line 2444   + 
at Kernel ->handle (object(Request)) 
in web/app_dev.php at line 28   + 
Logs   - 1 error
INFO - Matched route "claro_workspace_subscription_url_generate_anonymous". 
INFO - Populated SecurityContext with an anonymous Token 
INFO - The ability to pass file names to the Symfony\Component\Yaml\Yaml::parse method is deprecated since version 2.2 and will be removed in 3.0. Pass the YAML contents of the file instead. 
INFO - The ability to pass file names to the Symfony\Component\Yaml\Yaml::parse method is deprecated since version 2.2 and will be removed in 3.0. Pass the YAML contents of the file instead. 
INFO - The ability to pass file names to the Symfony\Component\Yaml\Yaml::parse method is deprecated since version 2.2 and will be removed in 3.0. Pass the YAML contents of the file instead. 
INFO - The ability to pass file names to the Symfony\Component\Yaml\Yaml::parse method is deprecated since version 2.2 and will be removed in 3.0. Pass the YAML contents of the file instead. 
INFO - No expression found; abstaining from voting.

dtisch:
Conversion from form data array to User entity resolves #741


Elorfin:
I think a cleaner way (not tested) would be to set the correct `data_class` (value needs to be `Claroline\CoreBundle\Entity\User`) in the definition of `BaseProfileType`.

With this property sets, `$form->getData()` will give you an instance of `User`.

dtisch:
As I didn't find another issue which seems to be related to #741, I decided not to modify `BaseProfileType` because I don't know if this form class is used somewhere else in the platform. Your cleaner solution works for #741 but I'm not 100% sure that no other part of code expects `$form->getData()` to be an array.

Elorfin:
As I can see it's only used here and for registration : https://github.com/claroline/Distribution/blob/617f4cbdc32b2b21db738cc21cbd120c1c8c1761/main/core/Controller/RegistrationController.php#L129.

The registration also expects a User (and get one because the Form is initialized with an empty User object).

A form used to edit/create an Entity should return this entity, not an array that you need to use to populate another Entity object. 

Maybe other @claroline/maintainers can give their advice on it.

Elorfin:
Closed in favor to #749.

PR #753, opened 2016-06-30
add css exclusion class to video-player mutation observer

pitrackster:
allow to exclude a video tag from video-player mutation observer and videojs appliance by adding a css class ('not-video-js')

### DIFF
--- a/plugin/video-player/Resources/modules/watcher/observe.js
+++ b/plugin/video-player/Resources/modules/watcher/observe.js
@@ -1,3 +1,4 @@
+import $ from 'jquery'
 export default function observe (selector, callback, containers = [document.body]) {
   window.MutationObserver = window.MutationObserver || window.WebKitMutationObserver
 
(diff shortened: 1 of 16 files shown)


Elorfin:
LGTM

stefk:
@pitrackster, you should be able to fix the build issue now that #757 is merged.

plugin/video-player/Resources/modules/watcher/observe.js
```diff
@@ -17,8 +19,10 @@ export default function observe (selector, callback, containers = [document.body
               keepGoing = false
             }
           })
+          // check if the element has the exclude class
+          let exclude = el.className.indexOf(excludeClass) > -1
```
kitan1982: Shouldn't you use "const" here instead of "let" ?

kitan1982:
LGTM

Elorfin:
LGTM

PR #754, opened 2016-06-30
[ExoBundle] Allow anonymous to pass an Exercise

Elorfin:
This is the first working version of the anonymous access (issue #74).

As discussed in meeting, this implementation is not perfect but it works almost the same way the PHP did.

The problems of the current implementation are : 
- If the User refreshes the page, a new Paper is generated (it's IMHO the biggest problem).
- If the User changes the paper ID in the URL, he can access papers of other anonymous (but not the ones of authenticated Users)
- The User can only see his results at the end of the Exercise. He will not be able to get them if he quit the correction page.
- Some features of the Exercise can not work (e.g. max attempts).

This problems need to be fixed for the 16.09 release, so I don't close the related issue.

Is that OK for you @claroline/icap-ec ? Is there a feature you miss in the implementation ?


stefk:
Why are you removing all these tests? AFAIK only one of them is concerned by your changes.

Elorfin:
I have removed the corresponding entry point in the api because they are not used anymore. So the tests are obsoletes. It's just code cleaning and it's not directly related to the feature.

stefk:
Most of the tested scenarios are still valid:

- submit an answer
- submit an answer in an invalid format
- submit an answer after the deadline
- submit an answer in another user paper

Those tests can't be obsolete. The features have problably been moved and tests and old code were left behind... It deserves a proper issue IMO.

Elorfin:
Yes. The tests for the new methods have not been added. I will add them in this PR

stefk:
LGTM

kitan1982:
LGTM

stefk:
LGTM

Issue #734, opened 2016-06-24
Load jQuery and jQuery UI from webpack


Elorfin:
For now it's not possible to include jquery and jquery UI with : 

```
import $ from 'jquery/dist/jquery'
import 'jquery-ui/jquery-ui'
```

As it creates a local `$` object which is different from the `window.$` which is available in the browser.

PR #757, opened 2016-06-30
Allow jquery and jquery-ui to be required through imports

stefk:
Related to #734. Ideally when writing js modules, we should require everything needed through imports statements. However, we are currently stuck with a global instance of jquery injected via a script tag. Rather than allowing the use of this implicit global, this PR makes it available through the module system, so that:
- modules that consume jquery are correctly written even if jquery is not really managed by webpack
- when we're ready to remove the script tag or if we find a better config, nothing must be changed in the modules themselves
- eslint is happy

The PR also allows jquery ui modules to be loaded correctly.

**TLDR**:

```js
import $ from 'jquery'
import 'jquery-ui/ui/draggable'
import 'jquery-ui/ui/droppable'

$('#foo').draggable()
$('.bar').droppable()
```

### DIFF
--- a/main/core/Resources/modules/jquery/index.js
+++ b/main/core/Resources/modules/jquery/index.js
@@ -1,0 +1,11 @@
+/**
+ * This module simply makes the global jQuery instance (injected via a script
+ * tag for compatibily reasons) available to other modules. It is aliased in the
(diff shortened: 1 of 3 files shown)


kitan1982:
LGTM

Elorfin:
LGTM

PR #759, opened 2016-07-01
[CoreBundle] Previous desktop refactoring

kitan1982:
- See https://github.com/claroline/Distribution/pull/585
- Reverts commit cfbf836


stefk:
@kitan1982, too large to review... It's probably better to spawn a docker instance for this one.

kitan1982:
http://pr-759-1467632369.travis-build.claroline.net/claroline/

ptsavdar:
I've tried to test most of the functionalities and here are my remarks:
- Recent activity widget won't show graph and would produce some errors.
- One cannot target a tab in desktop. That is there is no link associated directly with a tab. It would be nice if someone could target the desired tab directly instead of loading page and then choosing the tab (create different routes for tabs)

For the moment I think that's all. I'll come back if I notice anything else.
Thank you for your great work!

kitan1982:
Thank you for the test.

The graph in the "Recent activities" widget doesn't show because I have commented "jquery.jqplot.js" import in "main/core/Resources/views/Log/count_by_day_chart.html.twig".
Otherwise the desktop displays a blank page. I don't know why.
Also I think the errors come from the absence of "jquery.jqplot.js" 

For the possibility to target a tab, I have planned to do it later.

ptsavdar:
Ok, concerning the tabs, for the moment is ok but eventually we will need this functionnality back. However, if the same changes will be applied to workspaces then for us it's an absolute need to be able to target tabs but also that tab URLs will stay intact (that is the URL before angularisation stays the same as the one afterwards).

kitan1982:
Actually the same changes have already been applied to workspaces and the behavior is exactly the same.
I'm not very familiar with ui-router but I'm not sure it will be possible to keep the same URL.
I will check but I can't garantee a positive result.

kitan1982:
The best I could do is to keep the method in PHP (with the old URL) and make a redirection

ptsavdar:
Ok. The thing is for us is crucial that the URLs stay the same or at least work even after the changes. Otherwise we cannot update to this version. Many of our users have based their work on tabs in workspaces and shared those links with their learners/followers. So if the link won't work you can see how problematic and confusing this could be for them.

kitan1982:
Yes, it's understandable.
I will put that solution in place when I have time.

ptsavdar:
Great! Thank you @kitan1982 ! We will wait for this implementation before merging to master.

LaurentGruber:
LGTM

ngodfraind:
LGTM

PR #767, opened 2016-07-04
Merge branch 7.x

Elorfin:
FYI, there was no conflict to resolve.


purplefish32:
The Travis build is out of memory

Elorfin:
Yes we are checking this. The last commit increases the memory to run composer.

Elorfin:
@claroline/maintainers : we will need to bypass travis for this PR as the additionnel linters rules catch too many errors and it will be complicated to fix all of them (like the loose comparisons which can generate bugs if replaced).

PR #793, opened 2016-07-05
[ExoBundle] Stand-alone module for Feedbacks

Elorfin:
For now only the `FeedbackService` has been moved in.
This will permits to also put the retrieve of feedbacks for each question types (will make the `Question` module smaller).


ptsavdar:
LGTM

stefk:
LGTM

PR #794, opened 2016-07-05
Exo keep tries

eldoniel:
This PR fixes the following issue:
https://github.com/claroline/Distribution/issues/531

In formative exercises, the steps gets locked:
- when all tries are used
- when all answers are found

If the user changes step, and then goes back to an already played one, the step will stay locked.


ptsavdar:
```
console.log('Message: ' + msg)
console.log('Code: ' + code)
console.log('Status: ' + status)
```
I think is better to use the Angular [$log](https://docs.angularjs.org/api/ng/service/$log) service instead of the console.log (I think I saw it in the best practices or something that @stefk sent us the other day)

eldoniel:
There is still one syntax error I can't get rid of. I don't know how to import 'Routing' from fosjsbundle. Any ideas?

Thanks in advance.

ptsavdar:
Have you tried using window.Routing instead of just Routing? I'm not sure it will work but you can try it, if you haven't already tried that.

stefk:
@eldoniel, I don't think it's possible to make a clean import of that lib. It's available globally on every page anyway and we can't change that now, so the only problem is to tell eslint about that global.

I'm pretty sure I mentionned it in the chat, but the best solution IMO is to use a service defined in the core bundle that wraps the router (`main/core/fos-js-router`), so that the global [specification](http://eslint.org/docs/user-guide/configuring#specifying-globals) is done there and you don't have to refer to `Routing` directly.

eldoniel:
@stefk @ptsavdar Thank you for your answers.

I'm not sure I understood well what I had to do. I tried something anyway; it seems to be working and to pass tests better.

main/core/Resources/modules/fos-js-router/module.js
```diff
@@ -1,8 +1,7 @@
...
 import 'angular/angular.min'
 
+import RouterService from './Service/RouterService'
+
 angular.module('ui.fos-js-router', [])
-    .filter('path', function () {
-        return function (route, parameters = {}) {
-            return Routing.generate(route, parameters)
-        };
-    })
```
stefk: Are you sure no existing code use this filter?

main/core/Resources/modules/fos-js-router/Service/RouterService.js
```diff
@@ -0,0 +1,189 @@
+/* global goog */
```
stefk: Where does this code come from ?

plugin/exo/Resources/modules/paper/Services/UserPaperService.js
```diff
@@ -14,6 +14,8 @@ function UserPaperService($http, $q, PaperService, ExerciseService) {
   this.ExerciseService = ExerciseService
 }
 
+/* global Routing */
```
stefk: The point is to have that annotation in the router service, not here.

main/core/Resources/modules/fos-js-router/Service/RouterService.js
eldoniel: I did add it myself, as I was getting a "no-undef" error for this "goog" as well.
(sorry if I do weird things, all of this is really blurry for me right now, I'm a bit lost...)

plugin/exo/Manager/ExerciseManager.php
```diff
@@ -81,7 +81,7 @@ public function deleteStep(Exercise $exercise, Step $step)
 
         // Update steps order
         $steps = $exercise->getSteps();
-        foreach ($steps as $pos => $stepToReorder) {
+        foreach ($steps as $pos) {
```
Elorfin: In fact, there is a bug in this method : you need to keep the `$stepToReorder` variables and do `$stepToReorder->setOrder($pos)` (next line).

This method delete the step `$step` and reorder the others. So it's not the deleted one we need to reorder.

main/core/Resources/modules/fos-js-router/module.js
ptsavdar: I know I'm using it in website I think and also in portal.

plugin/exo/Manager/ExerciseManager.php
```diff
@@ -271,6 +271,10 @@ public function importExercise($data)
      */
     public function exportExercise(Exercise $exercise, $withSolutions = true)
     {
+        if ($exercise->getType() === '3') {
+            $withSolutions = true;
+        }
```
Elorfin: There are constants in `Exercise` entity for type. It's better to use it instead of the direct value.

Elorfin:
LGTM

ptsavdar:
LGTM

PR #795, opened 2016-07-05
Fixed bug with csv workspace import, added flashbox in users admin list

ptsavdar:
- Fixed bug in workspace users import with csv
- Added flashbox to users list in admin
- Changed path summary from default "pinned" to default "unpinned"

### DIFF
--- a/plugin/path/Resources/public/js/Path/Service/PathService.js
+++ b/plugin/path/Resources/public/js/Path/Service/PathService.js
@@ -37,7 +37,7 @@
              */
             var summary = {
                 opened: true,
-                pinned: true
+                pinned: false
             };
 
             /**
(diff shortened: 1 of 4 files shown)


plugin/path/Resources/public/js/Path/Service/PathService.js
```diff
@@ -37,7 +37,7 @@
              */
             var summary = {
                 opened: true,
-                pinned: true
+                pinned: false
```
ngodfraind: This is unrelated to the title of the issue ? Should it be merged aswell ?

ptsavdar: Yes, I know but to gain some time I pushed it here. It's just a minor detail.

ngodfraind:
LGTM

ptsavdar:
LGTM

PR #825, opened 2016-07-08
[PortfolioBundle] Fixed tinymce issue in portfolio

ptsavdar:


plugin/portfolio/Resources/views/Portfolio/index.html.twig
```diff
@@ -124,14 +124,12 @@
     <script type="text/javascript" src="{{ url('bazinga_jstranslation_js', { 'domain': 'icap_portfolio' }) }}"></script>
 
     {% javascripts debug=false filter='jsmin' output='js/portfolio/index/vendor.js'
-        '@StfalconTinymceBundle/Resources/public/vendor/tinymce/tinymce.jquery.min.js'
-        '@StfalconTinymceBundle/Resources/public/vendor/tinymce/jquery.tinymce.min.js'
         'packages/angular/angular.min.js'
         'packages/angular-resource/angular-resource.min.js'
         'packages/angular-sanitize/angular-sanitize.min.js'
         'packages/angular-animate/angular-animate.min.js'
         '@InnovaAngularUIResourcePickerBundle/Resources/public/js/angular-resource-picker-directive.js'
```
Elorfin: I know it's not part of your PR but can you also load this library from `bower` instead ?
There is no difference between the two implementations, so it's not too difficult and `InnovaAngularUIResourcePickerBundle` needs to be removed from the dependencies.

ptsavdar: Yeap! Of course! Thanks for the feedback! I'm doing this right away.

Elorfin:
LGTM

ptsavdar:
LGTM

PR #833, opened 2016-07-11
[ExoBundle] Fix on formative cloze questions

eldoniel:
This PR fixes the following issues:
https://github.com/claroline/Distribution/issues/807
https://github.com/claroline/Distribution/issues/809

On the editor, we can set multiple right answers for one hole, that will both be valid on the correction. But this case was not treated in the immediate feedbacks on formative exercise, in player mode. This PR fixes this.


stefk:
LGTM

Elorfin:
LGTM

PR #834, opened 2016-07-12
Merge branch 7.x

Elorfin:


Elorfin:
As the last merge of the 7.x, I have applied the new CS fixer rules but I can not easily fix the loose comparisons errors.

So we will need to disable Travis checks to merge this PR.

PR #844, opened 2016-07-13
Exo fixes pairing

eldoniel:
This PR fixes the following issue:

https://github.com/claroline/Distribution/issues/803

Reseted proposals couldn't be moved again. This bug gets fixed.

### DIFF
--- a/plugin/exo/Resources/modules/question/Controllers/Type/MatchQuestionCtrl.js
+++ b/plugin/exo/Resources/modules/question/Controllers/Type/MatchQuestionCtrl.js
@@ -484,9 +484,9 @@ MatchQuestionCtrl.prototype.reset = function reset() {
         $(".origin").each(function () {
             if ($(this).find('.draggable').attr('style')) {
                 $(this).find('.draggable').removeAttr('style');
-                $(this).find('.draggable').removeAttr('aria-disabled');
-                $(this).find('.draggable').draggable("enable");
             }
+            $(this).removeAttr('aria-disabled');
+            $(this).draggable("enable");
         });
 
         // init all drop containers ui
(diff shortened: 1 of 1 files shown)


kitan1982:
LGTM

stefk:
@eldoniel, can you point to the parts that actually changed? The fix is kinda lost in cs changes.

plugin/exo/Resources/modules/question/Controllers/Type/MatchQuestionCtrl.js
```diff
@@ -450,130 +451,130 @@ MatchQuestionCtrl.prototype.onFeedbackHide = function onFeedbackHide() {
...
-};
+  if (this.question.toBind) {
+    jsPlumb.detachEveryConnection()
+    this.connections.splice(0, this.connections.length)
+  } else {
+    // init all proposals ui
+    $('.origin').each(function () {
+      if ($(this).find('.draggable').attr('style')) {
+        $(this).find('.draggable').removeAttr('style')
+      }
```
eldoniel: @stefk Here is done the bugfix. The two lines below were in the if above.

plugin/exo/Resources/modules/question/Controllers/Type/MatchQuestionCtrl.js
```diff
@@ -450,130 +451,130 @@ MatchQuestionCtrl.prototype.onFeedbackHide = function onFeedbackHide() {
...
+      $(this).removeAttr('aria-disabled')
+      $(this).draggable('enable')
```
eldoniel: @stefk Here is done the bugfix. The two lines "removeAttr" and "draggable" were in the if above.

LaurentGruber:
LGTM

PR #848, opened 2016-07-13
Path conditions lock

CPASimUSante:
This is a PR for a new feature for path Resource : 
It allows, when conditions are set, that the student may ask for unlocking a step. The WS manager receives a notification, can unlock the step. The student then receives a notification that the step is unlocked and can access it.


CPASimUSante:
I don't know if this PR must be squashed before being merged since this is a feature ?

Elorfin:
Yes this needs to be squashed. For new features it's not really interesting to keep all the steps needed to implement it. And in most case, there are "wip" commits or some which break some things.

No squash are most likely for PR which fix multiple issues or which include multiple features.

Also, if you want to keep the history of a PR, you will need to avoid the merge commits in it and use `git rebase` instead (we can discuss about it in the chat if you want).

About this PR, there is a lot of code and it will be difficult to make a full code review on it. Can you spawn a docker instance so we will be able to test it ?

plugin/path/Event/Log/LogStepUnlockDoneEvent.php
```diff
@@ -0,0 +1,109 @@
+<?php
+
+namespace Innova\PathBundle\Event\Log;
+
+use Claroline\CoreBundle\Event\Log\AbstractLogResourceEvent;
+use Claroline\CoreBundle\Event\Log\NotifiableInterface;
+use Innova\PathBundle\Entity\Step;
+
+class LogStepUnlockDoneEvent extends AbstractLogResourceEvent //log associated to a resource
```
Elorfin: Weird place for a comment (and not really mandatory).

plugin/path/Event/Log/LogStepUnlockEvent.php
```diff
@@ -0,0 +1,107 @@
+<?php
+
+namespace Innova\PathBundle\Event\Log;
+
+use Claroline\CoreBundle\Event\Log\AbstractLogResourceEvent;
+use Claroline\CoreBundle\Event\Log\NotifiableInterface;
+use Innova\PathBundle\Entity\Step;
+
+class LogStepUnlockEvent extends AbstractLogResourceEvent //log associated to a resource
```
Elorfin: Same here.

Elorfin:
I have got a little check to the code and I have found some vars that are not written in camel case. That's the standard for PHP. I have found many, so I have not commented all in the code. But for example there is `$lockedcall` in the entity UserProgression which should be `$lockedCall` (idem for the corresponding getter and setter).

IMO, this is not to be fixed (this will require too many time), I just wanted to point out for your next PRs.

stefk:
Code wise, LGTM. But this needs a real test as @Elorfin said.

CPASimUSante:
I will correct as many things than i can (find). 
I'm ok for spawning a docker instance but i haven't noted how. I'll ask again in the chat.

Elorfin:
Closed in favor of #855.

PR #860, opened 2016-07-18
[ExoBundle] Improves player rendering

Elorfin:
- Add bootstrap classes in match questions for proper display on mobile screens.
- Increase size of img / video in match questions. it's 200px instead of 110px. (@Gatomlo I don't know if it will be sufficient for you, but I can't increase the size more without sacrifice rendering on small screens).

We will need to wait the new editor planned for the 16.09 for a better management of each resource type (for now, even the zoom feature of "to pair" questions doesn't work well with images).

### DIFF
--- a/plugin/exo/Resources/less/question/type/match.less
+++ b/plugin/exo/Resources/less/question/type/match.less
@@ -2,6 +2,8 @@
     position: relative;
 
     div > p {
+        margin-bottom: 0;
+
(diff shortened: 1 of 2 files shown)


LaurentGruber:
LGTM

kitan1982:
LGTM

LaurentGruber:
LGTM

PR #868, opened 2016-07-20
[Collecticiel]Issue867 : return receipt values

ericvincenterv:
fix claroline/Distribution#867 bug return receipt values


LaurentGruber:
LGTM

Elorfin:
LGTM

PR #878, opened 2016-07-28
[PathBundle] Fixes for resource iframe in PathBundle

dtisch:
- In some cases, the iframe content height can't be determined at the right time (with SCORM, for example). The loop will now ignore this situation.
- In some cases, an empty iframe (with no resource attached) occupies space on screen anyway. It is now hidden if empty.

### DIFF
--- a/main/core/Resources/views/Workspace/layout.html.twig
+++ b/main/core/Resources/views/Workspace/layout.html.twig
@@ -120,7 +120,7 @@
 
             setInterval(function() {
                 var newHeight = getIframeWindowHeight();
-                
+    
                 if (newHeight !== null && newHeight !== oldHeight) {
                     postHeight(newHeight);
                     oldHeight = newHeight;
--- a/main/core/Resources/views/Workspace/layout.html.twig
+++ b/main/core/Resources/views/Workspace/layout.html.twig
@@ -121,10 +121,8 @@
             setInterval(function() {
                 var newHeight = getIframeWindowHeight();
 
-                if (newHeight =! null) {
-                    if (newHeight != oldHeight) {
-                        postHeight(newHeight);
-                    }
+                if (newHeight !== null && newHeight !== oldHeight) {
+                    postHeight(newHeight);
                     oldHeight = newHeight;
                 }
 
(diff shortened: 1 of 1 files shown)


main/core/Resources/views/Workspace/layout.html.twig
```diff
@@ -113,10 +113,14 @@
 
             setInterval(function() {
                 var newHeight = getIframeWindowHeight();
-                if (newHeight != oldHeight) {
-                    postHeight(newHeight);
+
+                if (newHeight =! null) {
+                    if (newHeight != oldHeight) {
+                        postHeight(newHeight);
+                    }
```
Elorfin: If statements can be simplified : `newHeight =! null && newHeight != oldHeight`. Also you need to use strict comparison `!==`.

plugin/path/Resources/public/js/ResourcePrimary/Directive/ResourcesPrimaryShowDirective.js
```diff
@@ -11,7 +11,7 @@
                 replace: true,
                 controller: ResourcesPrimaryShowCtrl,
                 controllerAs: 'resourcesPrimaryShowCtrl',
-                template: '<iframe id="embeddedActivity" style="width: 100%; min-height: {{ resourcesPrimaryShowCtrl.height }}px;" data-ng-src="{{ resourcesPrimaryShowCtrl.resourceUrl.url }}" allowfullscreen></iframe>',
+                template: '<iframe id="embeddedActivity" data-ng-hide="resourcesPrimaryShowCtrl.resourceUrl.url == null" style="width: 100%; min-height: {{ resourcesPrimaryShowCtrl.height }}px;" data-ng-src="{{ resourcesPrimaryShowCtrl.resourceUrl.url }}" allowfullscreen></iframe>',
```
Elorfin: This needs to be strictly compared or you can use `data-ng-if="resourcesPrimaryShowCtrl.resourceUrl.url"`. NgIf will completely remove the tag from the DOM structure while NgHide only apply a `display: none` style on it. For states that not change very often, it's better to use NgIf to keep a cleaner DOM.

ptsavdar:
LGTM

Elorfin:
LGTM

PR #882, opened 2016-08-01
Update composer.json

ngodfraind:
Adding missing backslashes.


ngodfraind:
@GeoRoux 

It breaks packagist otherwise.

purplefish32:
LGTM

PR #900, opened 2016-08-16
[CoreBundle] Group import for users.

ngodfraind:
Quality of life improvements. When importing users, if a group is specified but doesn't exist yet, it'll be automatically created.


LaurentGruber:
LGTM

purplefish32:
LGTM

PR #901, opened 2016-08-16
Translations js

ngodfraind:
for https. I don't understand why all the other commits go here aswell


stefk:
Occurrences of `'js/translations/platform/*.js'` should be removed (that domain is already loaded in *Layout/javascripts.html.twig*).

> I don't understand why all the other commits go here aswell

You've probably started this branch from the `cursus` branch.

PR #910, opened 2016-08-18
[CoreBundle] hevinci_links opened in new tab.

ngodfraind:
This is a small improvements wich is sometimes requested. Resources links should be opened in a new tab.

This is the q&d fix but the resource manager is already a mess so I guess it's good enough for now.


LaurentGruber:
LGTM

stefk:
-1, specific plugin behaviour shouldn't be hardcoded in the core. There are other ways to achieve that, without adding yet another dirty trick in the resource manager.

Also:
- I'm not sure links should always be opened automatically in a new tab
- Virtually all resources could benefit from that feature

A solution could be to simply change the resource nodes markup and use proper `<a>` tags around the icons. Advantages:

- The resource manager would be lighter (no need to handle opening clicks in js)
- The markup would be more correct/semantic/accessible
- Users could open *any* resource in a new tab just by using the ctrl-click/"Open link in a new tab" feature of their browser.

ngodfraind:
I agree but this is just a quick fix for a rather common request. I did not intend to rewrite the resource manager.

PR #914, opened 2016-08-19
[PathBundle] Wrong entity path

ngodfraind:
I found this while looking for something else. This is not the correct entity full name.


Elorfin:
LGTM

purplefish32:
LGTM

PR #951, opened 2016-08-23
[CoreBundle] Import: file parsed for rich text.

### LINKED ISSUE #883: [Corebundle] Import : replace UID in src of image
At the export of claro v1, I can export all files (html, img,...) and replace src with UID present in the manifest but at the import in claroline connect, the UID present in the html files, aren't replace by the new url of the images.  
This works with the txt files but not the html files.

ngodfraind:
Fixes #883 

The code is very dirty so it must be enabled manually by setting 'enable_rich_text_file_import' to true in platform_options.

It will be cleaner when we'll use uid instead of what we have now.


ngodfraind:
I like how a simple merge made travis mad because of phpcs lol.

main/core/Library/Transfert/ConfigurationBuilders/Tools/Resources/FileImporter.php
```diff
@@ -111,18 +117,57 @@ public function export(Workspace $workspace, array &$_files, $object)
...
                 'path' => $uid,
                 'mime_type' => $object->getResourceNode()->getMimeType(),
-            )));
+            ]]];
         }
 
         return $data;
     }
 
+    public function format($data)
```
stefk: Where is it called?

stefk: Where does data come from? Almost impossible to follow what's below...

ngodfraind: $data is the object configuration in the yml file.
It usually looks like this:
```
item:
   name: Test_image.html
   creator: null
   parent: 0
   type: file
   roles:
        -
             role: { name: ROLE_WS_COLLABORATOR, rights: { open: true, export: true } }
    uid: 4
    data:
          -
               file: { path: 57a1a8a17766f.html, mime_type: text/html }
```
So what we do is that we get the html file, find what node is linked to the html file (wich is a pain because the "uid" isn't a guid yet so we have to parse each file to find wich one match properly; it's like a "findByFile('xxx') method and then we edit it.

It's going to be much more easier to understand when guid will be used everywhere.

stefk: You could add that explanation in a doc block, it would help future readers. Otherwise if we can't get something cleaner for now I'm good with it (once phpcs errors are fixed).

LaurentGruber:
LGTM

ngodfraind:
LGTM

PR #955, opened 2016-08-24
Fix sql type errors in tests

stefk:
Some facet tests are wrongly inserting strings in integer columns. How can most mysql versions let that happen by default is beyond me, but locally I now need this fix.


LaurentGruber:
LGTM

ngodfraind:
LGTM

PR #956, opened 2016-08-24
[ScormBundle] First version of SCORM export

Elorfin:
This PR adds a command to export a resource in a SCORM package.

This works with events like other resources action (create, open, copy, etc.).

The documentation can be found here : [README](https://github.com/claroline/Distribution/blob/c76c03bbcdcf11baf5b0f7a08da34c42cae03389/plugin/scorm/README.md). It misses an example, but a wip implementation can be found here : [Exo export](https://github.com/claroline/Distribution/blob/exo-scorm-export/plugin/exo/Listener/ExerciseListener.php#L190).

Any idea to improve the implementation is welcome.
The things I particularly dyslike for now are : 
- The use of `FOSJsRoutingBundle` controller inside a manager to generate the JS routes file. All the work is done in the controller, so I have not found other solution for now.
- The use of `Twig\WebpackExtension` to access the webpack compiled files. Maybe create a service to put the logic in and rely on it in the exporter and the twig extension.
- The use of the path of assets file in the `web` directory to add them into the archive.


stefk:
> The use of FOSJsRoutingBundle controller inside a manager to generate the JS routes file. All the work is done in the controller, so I have not found other solution for now.

Routes are dumped in *cache/[env]/fosJsRouting/data.json* but you'd have to wrap them in a 
`setData` call with other stuff like base url, prefix, host, etc., so I'm not sure it's worthwhile. 

> The use of Twig\WebpackExtension to access the webpack compiled files. Maybe create a service to put the logic in and rely on it in the exporter and the twig extension.

Only cleaner way to do it (but no big deal anyway).

> The use of the path of assets file in the web directory to add them into the archive.

Not a problem IMO.

LGTM

ngodfraind:
LGTM

Elorfin:
LGTM

PR #979, opened 2016-08-25
[CoreBundle] homePath SSL was wrong

ngodfraind:
I'll add an SSL parameter in the platform_options.yml later because the sf2 detection isn't always reliable (because proxies mess up everything).


stefk:
LGTM

purplefish32:
LGTM

purplefish32:
Sorry I clicked on the wrong button, this can be merged

PR #980, opened 2016-08-25
Attempt at fixing claroline/Claroline build

stefk:
Still doesn't work. Maybe the dependency cache is hiding a bug in a new version of some package. I deleted the cache, we'll see if we hit the same problem here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: php
 
 php:
-  - "5.5"
+  - "5.5.38"
 
 addons:
   apt:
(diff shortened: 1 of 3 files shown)


purplefish32:
LGTM

stefk:
@purplefish32, I'm trying to reproduce the build errors seen in claroline/Claroline, so it isn't good until it fails ;)

stefk:
lol

stefk:
Almost 100% sure #986 solves the build issues. I still can't reproduce the problem here though :(

PR #984, opened 2016-08-25
[CoreBundle] User disabling

ngodfraind:
- Users can be disabled in the administration
- a new 'is_removed' field exists to perform soft delete operations (wich were previously handled by the is_enabled field).

References #944


LaurentGruber:
LGTM

main/core/Repository/UserRepository.php
```diff
@@ -871,7 +871,7 @@ public function findEnabledUserById($userId)
             SELECT u
             FROM Claroline\CoreBundle\Entity\User u
             WHERE u.id = :userId
-            AND u.isEnabled = TRUE
+            AND u.isRemoved = TRUE
```
stefk: Shouldn't it be false instead?

main/core/Repository/UserRepository.php
```diff
@@ -885,7 +885,7 @@ public function findAllEnabledUsers($executeQuery = true)
         $dql = '
             SELECT u
             FROM Claroline\CoreBundle\Entity\User u
-            WHERE u.isEnabled = TRUE
+            WHERE u.isRemoved = TRUE
```
stefk: Same here.

main/core/Repository/UserRepository.php
ngodfraind: Good catch. I used some a search and replace but I did the first one manually and probably messed some one thel up.

ngodfraind:
I removed one the the methods (not used anywhere) and fixed the other one. It was used in a 4.xx  updater and in a debug command so it's pretty minor (and it probably could be removed aswell with some refactoring).

stefk:
LGTM

PR #985, opened 2016-08-25
[CoreBundle] Administration SSL flag.

ngodfraind:
SSL flag for https detection in case of proxies. Default is false.


main/core/Resources/views/layout.html.twig
```diff
@@ -25,7 +25,7 @@
 {% block layout %}
     <div id="sf-environement" data-env="{{ app.environment }}" class="hide"></div>
     <div id="homePath" class="hide">{{ path("claro_index") }}</div>
-    {% if app.request.isSecure() %}
+    {% if config.getParameter('ssl_enabled') %}
```
stefk: What happens if ssl is used but the parameter isn't set? Shouldn't it be `app.request.isSecure() or config.getParameter('ssl_enabled')` instead?

ngodfraind: It's would be better. I need to change 'ssl_enabled' to null by default then.

stefk: Why? It should work with a boolean too.

LaurentGruber:
LGTM

ptsavdar:
LGTM

PR #990, opened 2016-08-26
[CoreBundle] Fixing role add validation.

ngodfraind:
see lines  249 & 259

The abstract role subject already checks the role isn't already here when using the addRole() method (wich make that check redudant) and it caused some issues on a platform (and I don't know why yet).


main/core/Manager/RoleManager.php
```diff
@@ -246,25 +246,23 @@ public function associateRole(AbstractRoleSubject $ars, Role $role, $sendMail =
...
-        if (!$ars->hasRole($role->getName())) {
-            $ars->addRole($role);
-            $this->om->startFlushSuite();
+        $ars->addRole($role);
+        $this->om->startFlushSuite();
 
-            if ($dispatch) {
-                $this->dispatcher->dispatch(
+        if ($dispatch) {
+            $this->dispatcher->dispatch(
```
stefk: The event shouldn't be dispatched if the user already has the role.

ngodfraind:
@stefk 

About phpcs/phpmd, how can a valid PR become "wrong" after a merge ?

stefk:
@ngodfraind, that's because the last merge brought a fix in the travis config. Your changes weren't properly analysed before that.

ngodfraind:
The log is only done after a role is added but I'm not sure I want that (because if you add a role and fire this function, it won't always be logged).

LaurentGruber:
LGTM

stefk:
> if you add a role and fire this function

IMO that's a logic error. A simple way to prevent it could be to throw an exception in `AbstractRoleSubject::addRole` if the role is already present the collection. That way the api consumer is forced either to use the dedicated method of the manager or to take the responsability of persisting the role change directly.

ngodfraind:
It will have a lot of side effects with the current implementation unless we double check how roles are handled everywhere.

ngodfraind:
I'm ok with that but I think it's a different issue.

stefk:
> It will have a lot of side effects with the current implementation unless we double check how roles are handled everywhere.

Then the only correct way would be to check not only that the role is present with `hasRole` but that the role is in sync with the db. I see methods like `isDirty` or `getInsertDiff` in (http://www.doctrine-project.org/api/orm/2.3/class-Doctrine.ORM.PersistentCollection.html) that could do the trick. Otherwise we'll hit the problem you mentionned.

ptsavdar:
LGTM . It solves a problem that we hat in one of our platforms. The issue is that when we enroll a user in a group in which we have granted some platform roles (let's say ROLE_ADMIN or ROLE_WS_CREATOR etc.) and the user hadn't alone been attributed one of these roles then he is granted these roles through the group. Till here everything is fine. However, let's say we want to attribute one of these platform roles to user directly, though the profile edit form. We check these roles but after saving the user still doesn't have these roles. Worse, if the user already had one of these roles before, after any profile change he would not have the role. This is due to the method [associateRole](https://github.com/claroline/Distribution/blob/master/main/core/Manager/RoleManager.php#L239) in RoleManager and the control on line [249](https://github.com/claroline/Distribution/blob/master/main/core/Manager/RoleManager.php#L249) that checks if user already has the role, which in our case results to true since a group in which the user is enrolled grants it to him.

stefk:
@ptsavdar, thanks for your input. The only reason this PR isn't merged yet is https://github.com/claroline/Distribution/pull/990#issuecomment-244359917, but looking at the code again, it seems it's an independent issue, present in both versions. @ngodfraind ?

LaurentGruber:
LGTM

PR #992, opened 2016-08-26
[CoreBundle] Users are not shown the administration interface 

ngodfraind:
also remove the unused view.html.twig

the 'canEdit' boolean is used before showing the edit buttons (wich aren't going to work for a regular user because it's server protected).


LaurentGruber:
LGTM

stefk:
LGTM

PR #994, opened 2016-08-27
essPath improve

Elorfin:


main/core/Resources/views/Workspace/layout.html.twig
```diff
@@ -90,6 +90,28 @@
             window.Workspace.id = null;
         {% endif %}
     </script>
+
+    <script>
+        var postHeight = function postHeight() {
+            if (parent) {
+                console.log('child = ' + document.body.scrollHeight);
```
stefk: This log (and the other at line 107) is probably not necessary.

main/core/Resources/views/Workspace/layout.html.twig
```diff
@@ -90,6 +90,28 @@
...
+
+                var message = 'documentHeight:'+document.body.scrollHeight;
```
stefk: The message can be any sort of data, so an object might be a better fit here: `{ documentHeight: document.body.scrollHeight }`

main/core/Resources/views/Workspace/layout.html.twig
```diff
@@ -90,6 +90,28 @@
...
+
+                parent.postMessage(message, "*");
```
stefk: According to [mdn](https://developer.mozilla.org/fr/docs/Web/API/Window/postMessage), there's a security risk in using `*` as the targetOrigin parameter. If this listener is used for resources embedded in iframes, couldn't we use `document.origin` instead?

stefk:
There's commented code (js and less) and extra log statements in a several places, otherwise it looks good. But as it's a rather large pr with ui changes, it might be a good idea to start a docker instance to review it.

Elorfin:
Sorry. This was not supposed to be a new PR.
Some work in progress that have been integrated in other PRs.

PR #995, opened 2016-08-28
[PathBundle] Scorm export (+ webpack/ES6)

Elorfin:
This PR adds SCORM export for path. In order to make it work, the UI has been migrated to ES6/Webpack.

- implement Scorm export event listener
- migrate to ES6 / Webpack
- remove the `InnovaAngularUIPageslide` library from composer and load it from bower (next step is completly remove it)
- code refactoring and fixes of the conditions (needed to avoid AJAX requests on non conditional paths)

There are still some little bugs and problem with resources path that will need to be fixed in the next PR.

As the whole JS code base has changed, this is not reviewable, we will need to test it on one of our instance.

Also we will may need to disable travis for this PR as ESLint may not be happy and I will not be able to fix all its errors.


LaurentGruber:
LGTM

Elorfin:
ESLint has been fixed.

It also includes : 
- a `translation` module to expose Will Durand  JS translator as an Angular service (to avoid the direct use of the global object `Translator`)
- a controller and method to expose the resource icons the API (previously loaded through a global var populated in twig)

stefk:
Do we merge it as is or do we test it on a docker instance first? It looks good but as @Elorfin said, it's too large to be reviewed properly.

Elorfin:
We have planned to deploy it on one of our instance to test it.

The main feature here is the SCORM export and I don't think we can test it in a docker as it is a CLI command.

We will check for bugs in the same time.

Elorfin:
LGTM

PR #1002, opened 2016-08-30
Translation of Competency plugin

om64:
Complete


stefk:
@om64, existing quotes in translation files date from the time when they were required. Now, except in rare cases, they are completely superfluous. So if we're aiming at consistency and simplicity, removing them is probably better than adding new ones everywhere.

plugin/ldap/Resources/translations/ldap.fr.yml
```diff
@@ -1,28 +1,28 @@
...
-host: Host
-port: Port
-distinguished_name: DN
-username: Utilisateur
-password: Mot de passe
-protocol_version: Version
-append_dn: Ajouter le dn au nom d'utilisateur
-auto_creation: Créer l'utilisateur s'il n'existe pas
+Add server: 'Ajouter un serveur'
+Allow account creation at the first login: 'Permettre la création d'un compte lors de la première connexion'
```
Elorfin: Missing escape quote `d'un`

om64:
@stefk, What about a case such as "Ce placeholder sera remplacé par l'adresse de l'organisation" where I already have two separate quotes? It seems that editors recognize that kind of text quite better with quotes.

stefk:
@om64, the expression:

```yaml
msg: Ce placeholder sera remplacé par l'adresse de l'organisation
```

is valid in yaml. Most editors I know of understand it correctly (atom, webstorm, gedit, etc.). If they don't, it's just a bug in their syntax coloring algorithm that should be reported. I don't think we should type and read more characters to make those editors happy ;)

om64:
@stefk OK to remove quotes, but the fact is that I've added them to more than 100 files so far. What a waste of time...

stefk:
@om64, don't waste your time again in removing them, that's ok. We can do that later or even write a script to do it automatically if we agree that's better.

LaurentGruber:
LGTM

ngodfraind:
LGTM

PR #1005, opened 2016-08-31
[CoreBundle] Workspace import/export

ngodfraind:
It can be squashed.

The import/export workspace model command can be used to du bulk operations (exporting every ws as model, then importing every workspace contained in a directory).

@todo I forgot to edit the claro_media route. It should be able en handle guid because I generated some urls with them to make things simpler.

edit: guid are already handled. I just tested it.


LaurentGruber:
LGTM

kitan1982:
LGTM

PR #1015, opened 2016-08-31
[CoreBundle] event soft delete

ngodfraind:
$event isn't always defined so this is the monkey patch.

Maybe the $node->setActive() block should be in the `if ($node->getClass() !== 'Claroline\CoreBundle\Entity\Resource\ResourceShortcut') {`condition. I don't know.

### DIFF
--- a/main/core/Manager/ResourceManager.php
+++ b/main/core/Manager/ResourceManager.php
@@ -929,12 +929,11 @@ class ResourceManager
         $count = count($nodes);
         $nodes[] = $resourceNode;
         $softDelete = $this->platformConfigHandler->getParameter('resource_soft_delete');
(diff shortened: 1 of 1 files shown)


stefk:
There's something wrong in that code. If `$eventSoftDelete` is put outside the loop, its value when the node is a shortcut could depend on the value of the previous iteration, which is completety random. Maybe the shortcut logic should be fixed instead. Aren't you the one who wrote it?

ngodfraind:
You're right it should be in the loop and now it's ok. The shortcut logic is still a bit weird imo.

LaurentGruber:
LGTM

stefk:
LGTM

PR #1030, opened 2016-09-01
[CursusBundle] Cursus interpolation in expression

ngodfraind:
There is still the previous commit here but w/e.


ngodfraind:
It does 2 things:
- remove the interpolation in angular expression (it causes issue if `'` are present)
- force the return of the group list as an array (I don't know why the default behaviour was different, I also should have returned the count($groups) but it's not very important currently).

ptsavdar:
lgtm

LaurentGruber:
LGTM

PR #1034, opened 2016-09-01
[PathBundle] manages deleted resources + add id in structure

Elorfin:
This PR attempts to fix 2 bugs in Path : 
- The last version of the Angular UI get the ID of the path from the JSON structure, but for old path this property was not set.
- The copy didn't check if the resources (primary & secondary) exist, so if one is deleted there is an error. Now, the deleted resource is deleted in the structure of the copy.


ptsavdar:
LGTM

LaurentGruber:
LGTM

PR #1046, opened 2016-09-02
[ExoBundle] Fix mutliple issues

eldoniel:
Various urgent fixes on exo bundle:

- "Reset" button on match questions fixed
- remove visible html tags on additional fields
- enlarge iframes for "objects of the question"

### DIFF
--- a/plugin/exo/Resources/modules/question/Controllers/Type/MatchQuestionCtrl.js
+++ b/plugin/exo/Resources/modules/question/Controllers/Type/MatchQuestionCtrl.js
@@ -478,10 +478,7 @@ MatchQuestionCtrl.prototype.updateStudentData = function () {
  */
 MatchQuestionCtrl.prototype.reset = function reset() {
   if (this.question.toBind) {
-    var retrievedConnections = jsPlumb.getConnections()
-    for (var i = 0; i < retrievedConnections.length; i++) {
-      jsPlumb.detach(retrievedConnections[i])
-    }
+    jsPlumb.detachEveryConnection()
     this.connections.splice(0, this.connections.length)
   } else {
     // init all proposals ui
--- a/plugin/exo/Resources/modules/question/Controllers/Type/MatchQuestionCtrl.js
+++ b/plugin/exo/Resources/modules/question/Controllers/Type/MatchQuestionCtrl.js
@@ -478,7 +478,10 @@ MatchQuestionCtrl.prototype.updateStudentData = function () {
  */
 MatchQuestionCtrl.prototype.reset = function reset() {
   if (this.question.toBind) {
-    jsPlumb.detachEveryConnection()
+    var retrievedConnections = jsPlumb.getConnections()
+    for (var i = 0; i < retrievedConnections.length; i++) {
+      jsPlumb.detach(retrievedConnections[i])
+    }
     this.connections.splice(0, this.connections.length)
   } else {
     // init all proposals ui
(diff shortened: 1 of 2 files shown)


plugin/exo/Resources/modules/question/Controllers/Type/MatchQuestionCtrl.js
```diff
@@ -478,7 +478,10 @@ MatchQuestionCtrl.prototype.updateStudentData = function () {
  */
 MatchQuestionCtrl.prototype.reset = function reset() {
   if (this.question.toBind) {
-    jsPlumb.detachEveryConnection()
+    var retrievedConnections = jsPlumb.getConnections()
+    for (var i = 0; i < retrievedConnections.length; i++) {
+      jsPlumb.detach(retrievedConnections[i])
+    }
```
Elorfin: That's not the correct way to fix it. 
There is an option on the Endpoint that permits to use the `detachEveryConnection()`. This will also fix possible bugs on other call to this function (and there are).

Please revert this, I have the correct fix in my current branch.

plugin/exo/Resources/modules/question/Partials/show.html
```diff
@@ -43,13 +43,13 @@ <h4 class="panel-title">
...
             <!-- Description of the Question -->
-            <div data-ng-bind="questionShowCtrl.question.description"></div>
+            <div data-ng-bind-html="questionShowCtrl.question.description | unsafe"></div>
         </div>
         <div class="question-info" data-ng-if="questionShowCtrl.question.objects.length > 0">
             <h5>{{ 'question_object'|trans:{}:'ujm_sequence' }}</h5>
             <div data-ng-repeat="object in questionShowCtrl.question.objects">
-                <iframe src="{{ object.url }}"></iframe>
-                <span data-ng-if="object.type === 'text'">{{ object.data }}</span>
+                <iframe src="{{ object.url }}" style="height: 550px; width: 750px; resize: both; overflow: scroll;"></iframe>
```
Elorfin: An height of 550px seems a little big. It will take half of the screen for many users even if there is only 2 or 3 lines to display.

Also a fixed width will not be responsive. Use % instead or bootstrap grid.

Elorfin: It also misses a `ngIf` on the iframe tag to avoid empty IFrame (and 404 server call) in case there is no URL to resolve (which will result of a blank block of 550px with your fix).

plugin/exo/Resources/modules/question/Partials/show.html
```diff
@@ -106,12 +106,12 @@ <h5 data-ng-if="questionShowCtrl.includeCorrection && ((questionShowCtrl.questio
         
         <div class="question-info" data-ng-if="questionShowCtrl.question.specification">
             <h5>{{ 'question_specification'|trans:{}:'ujm_sequence' }}</h5>
-            <div>{{ questionShowCtrl.question.specification }}</div>
+            <div data-ng-bind-html="questionShowCtrl.question.specification | unsafe"></div>
```
Elorfin: Add `ngIf` for cleaner DOM.

LaurentGruber:
@eldoniel il y a des tickets associés aux changements qu'il faudrait lier ?

eldoniel:
@LaurentGruber Je ne crois pas. A la suite des commentaires d'Axel, j'ai annulé mes changements sur le bouton "reset" des match question (qui a un ticket associé), mais je ne crois pas avoir vu de tickets sur les changements restants, à savoir les balises html des champs supplémentaires et les tailles d'iframes.

LaurentGruber:
oki LGTM

ngodfraind:
LGTM

PR #1062, opened 2016-09-03
[ExoBundle] Initializes SCORM export and bug fixes

Elorfin:
This PR fixes a bunch of bugs : 
- adds some classes for better small screen display
- fixes confirm modal display if we use HTML content
- adds missing translations
- fixes reset button for association questions

It also : 
- initializes SCORM export for Exercise
- implements a stand alone for Hints (new angular module + specific API classes)
- avoids server calls if Hints are already loaded in the UI

**NB :** the export of Exercise is not yet fully functionnal, but if I make all the modifications needed here, the PR will not be reviewable. So I've decided to break it in smaller parts.

### DIFF
--- a/main/core/Resources/modules/modal/Partial/confirm-modal.html
+++ b/main/core/Resources/modules/modal/Partial/confirm-modal.html
@@ -1,6 +1,4 @@
-<div class="modal-body">
-    {{ confirmCtrl.content }}
-</div>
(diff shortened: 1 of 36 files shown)


LaurentGruber:
LGTM

purplefish32:
LGTM

PR #1070, opened 2016-09-05
[CoreBundle] Fixed jqplot issues on widget and logs

ptsavdar:
Basically instead of loading jqplot main file through jquery.jqplot.js, I created a dedicated twig file and load all necessary scripts in there. Otherwise, jquery.jqplot loaded those scripts dynamically and when it tried to insert them in the document using document.write it basically erased all document's content. Result = empty white page.


kitan1982:
LGTM

ptsavdar:
LGTM

PR #1072, opened 2016-09-05
[PathBundle] various fixes

### LINKED ISSUE #1066: [PathBundle] Saving in path editor
@Elorfin 
To activate/be able to click on the save button in the path editor, you have to write something in the path_decription.

Elorfin:
- Fixes bad condition in the copy if there are deleted resources in the path
- Disables delete button for root step
- Fixes #1066

### DIFF
--- a/plugin/path/Resources/modules/summary/Partial/edit-item.html
+++ b/plugin/path/Resources/modules/summary/Partial/edit-item.html
@@ -44,7 +44,7 @@
             </a>
 
             <!-- Delete step -->
-            <a href="" role="button" class="btn btn-danger btn-sm" data-ng-click="summaryItemEditCtrl.removeStep()" data-toggle="tooltip" data-container="body" title="{{ 'step_delete' | trans:{}:'path_wizards' }}">
+            <a href="" role="button" class="btn btn-danger btn-sm" data-ng-click="summaryItemEditCtrl.removeStep()" data-ng-disabled="0 === summaryItemEditCtrl.step.lvl" data-toggle="tooltip" data-container="body" title="{{ 'step_delete' | trans:{}:'path_wizards' }}">
                 <span class="fa fa-trash-o"></span>
                 <span class="sr-only">{{ 'step_delete' | trans:{}:'path_wizards' }}</span>
             </a>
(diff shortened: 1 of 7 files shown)


plugin/path/Resources/modules/summary/Partial/edit-item.html
```diff
@@ -44,7 +44,7 @@
             </a>
 
             <!-- Delete step -->
-            <a href="" role="button" class="btn btn-danger btn-sm" data-ng-click="summaryItemEditCtrl.removeStep()" data-toggle="tooltip" data-container="body" title="{{ 'step_delete' | trans:{}:'path_wizards' }}">
+            <a href="" role="button" class="btn btn-danger btn-sm" data-ng-click="summaryItemEditCtrl.removeStep()" data-ng-disabled="0 === summaryItemEditCtrl.step.lvl" data-toggle="tooltip" data-container="body" title="{{ 'step_delete' | trans:{}:'path_wizards' }}">
```
ptsavdar: I think that the ng-disabled despite the fact that it renders the anchor "inactive" it won't disable the "click" action. So basically, even though the button will be shown as inactive, essentially it still remains active (clickable).
See [here](https://plnkr.co/edit/rb8tBcyO8nEXJNCkZaip?p=preview) a plunk that I've created for the case. If you change the anchor to a button then it works fine.
I don't know if the above strange angularjs behavior is an absolute MUST NOT for your case or you can live with that, nevertheless I wanted to mention it just in case.

ptsavdar:
Besides my remark LGTM

plugin/path/Resources/modules/summary/Partial/edit-item.html
Elorfin: Thanks for pointing out. I will check that.

Elorfin:
@ptsavdar : do not merge the master branch when there is the `no-squash` tag. I make rebase to have clean history, so there will be possible conflicts if there are merge commits.

ptsavdar:
@Elorfin Oops! Sorry! I acted too fast. Forgot about the "no-squash".

Elorfin:
No worry. I will revert it if I can or squash the PR as the history is not so important here.

LaurentGruber:
LGTM

plugin/path/Resources/modules/summary/Partial/edit-item.html
Elorfin: I've added an additional check in the delete method to really disable the feature.

Elorfin:
LGTM

PR #1073, opened 2016-09-05
[CursusBundle] Cursus registration

ngodfraind:
This listener doesn't wait users are created to performs database queries (so if they're created in a batch, the query will crash).

Now we can force the session/cursus registration and we do it if the user has no id (there is no point querying anything in that case).

This change was suggested by @kitan1982 

We can also return null from the repositories if user->getId() is null (it may be better to do in the repository).


ptsavdar:
LGTM

kitan1982:
LGTM

ngodfraind:
I should remove the UserManager from the constructor. It's not needed anymore.

LaurentGruber:
LGTM

ngodfraind:
LGTM

PR #1076, opened 2016-09-05
[CoreBundle] resource copy less string

ngodfraind:
In production, copying a broken resource will just do nothing and return null with this fix.

I had an issue where a document from a dropzone/collecticiel wasn't removed properly and broke the workspace creation through a model (but copying the directory containing the resource would have the same issue).


LaurentGruber:
LGTM

purplefish32:
LGTM

Issue #238, opened 2016-04-29
[ExoBundle] Managing score


Elorfin:
In the current version of the ExoBundle, you can specify a global score for a Choice question : if the User respond correctly, we assign a score and if it's false, we assign another. The "regular" way is to assign  a score to each choice and give the student the point if this choice is selected.

The problem is this not managed (or partially) outside of the question editor : the api, the json quiz schema and the player can not handle it. Another problem is that the unit tests fail on this point because we can not import question which use this notation rule (the imported data are incomplete) and the total score of the question is not calculable.

As this point involves managing a new type of notation, we need to create a system which will permits to respond to the @claroline/icap-ec requirements (they have many notation systems to implement), even if we don't implement all the rules by now.

We have discussed about a possible implementation with @stefk (thanks ;)).

The idea is to add a `score` property in the Question object.

```
"score": {
  "type": "sum"
}
```

The `score` is an `Object` with a required property `type` which will define the notation rule for this question. So we will be able to make the correct calculation based on the score `type`. The `score` object will also define relevant property for the score `type`.

So a possible format for the global score could be : 

```
"score": { 
  "type": "fixed", 
  "threshold": 2, 
  "success": 10, 
  "failure": 3 
}
```

`success` : The score the student gets if his answer is right
`failure` : The score the student gets if his answer is false
`threshold` : The threshold from which the answer is considered right. @stefk if I have correctly understood the purpose of this property it's only relevant for question which have multiple reponses (like, multiple Choices, Holes, Match).

For the time being, we need only two rules : 
- type: sum. Currently used by Choice, Match, Graphic, Hole
- type: fixed. Currently used by Open, Choice (not implemented). We might also want to use it for Match, Hole and Graphic

To implement this, we have to : 
- [x] Upgrade the json-quiz schema to validate and handle this new property.
- [ ] Change the API mark workflow to handle multiple score calculation rules.
- [ ] Change the API to export this info into JSON.
- [ ] Check the Choice question import and modify it to handle global score.
- [ ] Update Tests suite (update existing + write new one for this particular point)

I open this issue to have your feedback and improve the implementation if you think this will not be sufficient to manage all our cases. We can discuss it in the team meeting.

I think this an important point to solve quickly : the editor permits to add global score for Choice, so somebody will rapidly find the bug.

@claroline/maintainers 
@claroline/ujm 
@claroline/icap-ec 
@claroline/innova

david42:
[...]

PR #1092, opened 2016-09-06
[ExoBundle] No paper save and angular scores

### LINKED ISSUE #529: [ExoBundle] Résultats
A répartir en plusieurs tickets ? Merci dans ce cas et désolé

J'ai droit à 5 tentatives pour cet exercice et j'en ai fait en réalité 4. Le tableau ne semble donc pas correct mais ceci est peut-être dû au bug que j'ai rencontré (cf. ticket traitant du dysfonctionnement de l'analyse des réponses dans le cas où l'on attribue un point positif à une réponse incorrecte)

Les heures indiquées sont incorrectes (il est +/- 20H)

Je pense que le score devrait avoir pour total le maximum de points, la somme des points (suggestion)

Suggestions : 
- utiliser le terme tentative pour les exercice et réserver essai pour les étapes
- colonne Interrompue peu compréhensible. Je verrais plutôt les choses à l'inverse : Terminé ? et changement d'icône en conséquence

Amélioration : 
- Ajouter une infobulle pour l’œil de la colonne des actions
- Ajouter une infobulle pour le V et X de la colonne interrompue

image

J'ai cliqué sur l'oeil de la tentative n°4 du tableau ci-dessus. Je m'attends à avoir accès directement à cette tentative, avec les réponses fournies, mais j'obtiens ceci. Est-ce normal ?

Comment obtient-on les réponses à un exercice déjà réalisé/terminé ?

Ceci dit les scores ne correspondent pas (20/20 dans le tableau et 10/20 ici) : 

Modifier Correction 5 par Tentative n°5. Voir suggestion ci-dessus, à tout le moins que les termes actuels (essai et correction) soient uniformisés

image

### LINKED ISSUE #532: [ExoBundle] Score-Etape
Le score n'est affiché apparemment que juste après la validation d'une étape
Pas dans les autres situations (retour en arrière par exemple) : 

image

### LINKED ISSUE #560: [ExoBundle] Erreur de score étape dans question ouverte mot
Il me semble que le score de la première question devrait être de 2/2 ou alors c'est l'analyse des réponses qui est incorrecte : 

image

### LINKED ISSUE #796: [ExoBundle] Player - Interrompre un exercice (amélioration et bug)
Situation : Si on a terminé la (seule ici) question de (la seule) étape et que l'on clique sur Interrompre, l’exercice est marqué Interrompu dans la liste des résultats. 

- [ ] Bug 1 : le score de la liste des résultats est incorrect (j'ai terminé et obtenu 2/2)
- [ ] Bug 2 : Pour terminer cet exercice, il faut cliquer sur "Recommencer l'exercice". Je me retrouve alors dans la question, réponses correctes déjà indiquées (j'avais terminé), score = 0/2 (j'avais obtenu 2/2), bouton valider l'étape actif (il ne devrait pas l'être). Pour terminer, je dois valider ce que je vois
- [ ] Amélioration 1 : Quand on a terminé toutes les étapes, on ne devrait pas pouvoir cliquer sur Interrompre, seulement sur Terminer
- [ ] Amélioration 2 : Quand un exercice est interrompu, on devrait pouvoir le poursuivre autrement qu'en cliquant sur "Recommencer l'exercice" ou en modifiant l'intitulé par "Poursuivre l'exercice"

image

### LINKED ISSUE #1020: [ExoBundle] {{ questionShowCtrl.question.title }}
Situation : sommatif, non publié, une seule question (association), test de l'exercice
Le bug se présente également après publication

Cf. {{ questionShowCtrl.question.title }}

### LINKED ISSUE #1025: [ExoBundle] Correction association : perte de sa réponse
Situation : sommatif, non publié, une seule question (association), test de l'exercice

Pour le cas ci-dessous : 
- non affichage de "sa" réponse lors de la correction (cf. Votre réponse)
- ne pas donner la possibilité d'indiquer/modifier sa réponse (cf. Votre réponse)

### LINKED ISSUE #1088: [ExoBundle] Bug analyse réponse version formative
Voir ci-dessous. la réponse est incorrecte. Bug dans l'analyse de la réponse. Mauvais feedback générale et on ne peut refaire la question alors que essai 1/5. Score correct et feedback spécifique ok

image

### LINKED ISSUE #1090: [ExoBundle] Description d'une étape
L'éditeur permet de faire ce qu'on fait partout dans l'éditeur riche mais, à l'affichage

- limité à 240 caractères
- pas d'affichage des ressources associées, dont image, vidéo, audio...

Elorfin:
This PR introduces two new features : 
- The possibility to avoid saving papers on the server (implemented with a flag in the `PaperService`). It will be possible to set it when an administrator try the exercise to avoid test papers in results.
- The calculation of scores in the front end.

**Score calculation :**
The server score has the priority. The new calculation system do not replace the old one. It is only used for formative exercises or if the paper is not saved. 

It is implemented this way  for performances reasons (we don't want to recalculate all papers scores in the list while the info is sent by the server for example) and to avoid possible problems with Papers which have questions that are no longer in the exercise (I think there will be no problem but it needs tests).

The system will need some code refactoring to fully implement #238. As the API do not send the correct info, this represents too many work to be done here.

**Bug fixes :** 
As this features have required many modifications, this PR also fixes the following problems (mainly score bugs) : 
- Lock match question edition when correction or feedback are shown (I think there is an issue for this, but I can't find it).
- Fixes #1025 
- Fixes #1020 
- Fixes #796 
- Fixes #532 
- Fixes #529 (the part about the score, the other parts have been done in another PR)
- Fixes #1088 
- Fixes #560 
- Fixes #1090 

Sorry for the big PR, but all the modifications are linked together so it's difficult to break in multiple PRs.


LaurentGruber:
LGTM

main/core/Resources/modules/utilities/Service/ArrayService.js
```diff
@@ -0,0 +1,48 @@
...
+    const picked = []
+    for (let i = 0; i < array.length; i++) {
+      if (i <= numberItems) {
+        picked.push(array[i])
+      }
+    }
+
+    return picked
+  }
+}
```
stefk: Not a problem per se, but I think we should consider using lodash for that kind of things:

https://lodash.com/docs/4.15.0#shuffle
https://lodash.com/docs/4.15.0#take

Elorfin: I must admit I have not checked if a library does it. 
Is lodash used a lot in the platform or will it be removed from dependencies (it is not in the mandatory list [here](https://github.com/claroline/Distribution/blob/master/doc/libraries/front-end.md#mandatory) ) ?

stefk: No, it's only an implicit dependency for now (although underscore has the same methods). But I think it would be wise to make it explicit and rely on it for such low level functions. We'd have a correctness/performance guarantee without having to test or document anything. Ping @claroline/commiters

plugin/exo/Resources/modules/paper/Services/UserPaperService.js
```diff
@@ -1,382 +1,455 @@
...
+        .error(() => {
+          for (let i = 0; i < stepPapers.length; i++) {
+            stepPapers[i].nbTries -= 1
+          }
+
+          deferred.reject([])
+        })
+    } else {
+      const exercise = this.ExerciseService.getExercise()
+      if ('3' === exercise.meta.type) {
```
stefk: `'3'` is really cryptic. Can we use constants to represent types instead?

plugin/exo/Resources/modules/paper/Services/UserPaperService.js
```diff
@@ -1,382 +1,455 @@
...
+    if (this.ExerciseService.isEditEnabled()) {
+      // Always show correction for exercise's administrators
+      available = true
+    } else {
+      // Use the configuration of the Exercise to know if it's available
+      const exercise = this.ExerciseService.getExercise()
+
+      switch (exercise.meta.correctionMode) {
+        // At the end of assessment
+      case '1': {
```
stefk: Same here and below, constants will make comments superfuous everywhere they appear in the codebase.

plugin/exo/Resources/modules/question/Services/QuestionService.js
```diff
@@ -1,60 +1,84 @@
...
+  getQuestion(questions, id) {
+    let question = null
+    for (let i = 0; i < questions.length; i++) {
+      if (id === questions[i].id) {
+        question = questions[i]
+        break // Stop searching
+      }
+    }
+
+    return question
```
stefk: Just a note for later: there's a lot of places where the code could be shortened by at least 60% using functional constructs like `forEach`, `map`, `reduce` and so on. Here the difference is striking: 

```js
getQuestion(questions, id) {
  return questions.find(question => question.id === id)
}
```

At this point a dedicated method is almost not needed at all.

plugin/exo/Transfer/Json/QuestionHandler/OpenHandler.php
```diff
@@ -124,7 +124,34 @@ public function convertInteractionDetails(Question $question, \stdClass $exportD
         $repo = $this->om->getRepository('UJMExoBundle:InteractionOpen');
         $openQuestion = $repo->findOneBy(['question' => $question]);
 
-        $exportData->scoreTotal = 5;
+        $exportData->typeOpen = $openQuestion->getTypeOpenQuestion()->getValue();
+
+        if ('long' === $exportData->typeOpen) {
```
stefk: Same remark about constants.

Elorfin: In fact this has to be removed as the total score of a question is calculated in Angular now from the question solution.

stefk:
There's room for small improvements here and there, but nothing blocking. So if it's needed now, LGTM.

Elorfin:
@stefk : I will add class constants before merge. For the other parts (like lodash and code simplification) we will wait for a code refactoring of the player.

main/core/Resources/modules/utilities/Service/ArrayService.js
ptsavdar: I backup Stephane on this. I think lodash should be one of our explicit dependencies. It has proven itself quite clean and quick. And with te "[lazy](http://filimanjaro.com/blog/2014/introducing-lazy-evaluation/)" syntax, I don't think we can ask for something better

ptsavdar:
LGTM

main/core/Resources/modules/utilities/Service/ArrayService.js
stefk: @ptsavdar, thanks for the link ;)

Elorfin: @ptsavdar : interesting article. thanks. I'm ok with both of you to add lodash in explicit dependencies and use it where it's useful.

Elorfin:
LGTM

Issue #840, opened 2016-07-12
[ExoBundle] Allow display the correction without score


Elorfin:
For now the only way to hide the score is to also hide the correction to users.
We need to be able to hide the score and still display correction.

ping @Gatomlo.

PR #1107, opened 2016-09-08
[ExoBundle] Allow display the correction without score

pitrackster:
New parameter that allow to hide all score informations in exercise


plugin/exo/Resources/modules/question/Partials/show.html
```diff
@@ -120,31 +121,36 @@ <h5 data-ng-if="questionShowCtrl.includeCorrection && ((questionShowCtrl.questio
                 <!-- Choice Type -->
                 <choice-correction
                         data-ng-switch-when="application/x.choice+json"
-                        data-question="questionShowCtrl.question">
+                        data-question="questionShowCtrl.question",
```
purplefish32: Are you sure this line needs a comma at the end ?
Seems like a mistake to me

pitrackster: yep you are right ! thx @purplefish32 !

LaurentGruber:
LGTM

plugin/exo/Resources/modules/paper/Controllers/PaperShowCtrl.js
```diff
@@ -38,6 +39,10 @@ PaperShowCtrl.prototype.showMinimalCorrection = function() {
   return this.PaperService.getExerciseMeta().minimalCorrection
 }
 
+PaperShowCtrl.prototype.hideScore = function() {
+  return this.PaperService.getExerciseMeta().markMode === this.UserPaperService.MarkMode.NEVER
```
Elorfin: It would have been better to create a service method for this in PaperService or UserPaperService.

Elorfin:
Related to #840

Elorfin:
LGTM

Issue #1108, opened 2016-09-08
[MediaResource] - Link help is to restrictive


pitrackster:
Checking [www] in the url is not necessary and blocking in some cases

pitrackster:
resolved by https://github.com/claroline/Distribution/pull/1112

PR #1112, opened 2016-09-08
help link pattern check for http(s):// prefix only

pitrackster:
Remove to restrictive constraints from media resource help link
Just check for http(s):// prefix
solve #1108


ptsavdar:
LGTM

LaurentGruber:
LGTM

PR #1116, opened 2016-09-09
[SurveyBundle] Add simple text and textarea without editor as answer fields

CPASimUSante:
Sofar in the Survey ressource there are 3 kinds of answer fields : 
- open with rich text editor
- multiple choice (unique and multiple answer)

This PR adds 
- the simple text field
- the simple textarea

as an answer field.
So this is not a new feature as such : it simply allows to display fields in a more appropriate way in some cases. When an expected anwser is a single word or short sentence ("What is your favourite dish ?") a textarea with  editor is maybe overkill.

PS: there's a Docker build to test it http://pr-1116-1473434584.travis-build.claroline.net/


plugin/survey/Controller/SurveyController.php
```diff
@@ -963,8 +968,13 @@ public function typedQuestionDisplayAction(Survey $survey, Question $question)
 
                 return $this->displayMultipleChoiceQuestion($question);
             case 'open_ended':
+            case 'open_ended_bare':
 
-                return $this->displayOpenEndedQuestion($question);
+                $isRich = ($questionType === 'open_ended') ? true : false;
```
stefk: Can be simplified to:

```php
$isRich = $questionType === 'open_ended';
```

plugin/survey/Controller/SurveyController.php
```diff
@@ -2060,10 +2153,19 @@ private function displayTypedQuestion(
                     $canEdit
                 );
             case 'open_ended':
-
+            case 'open_ended_bare':
+                $isRich = ($questionType == 'open_ended') ? true : false;
```
stefk: Same

plugin/survey/Controller/SurveyController.php
```diff
@@ -2143,6 +2246,26 @@ private function displayOpenEndedQuestion(
...
+            )
+        );
+    }
+
+    private function displaySimpleTextQuestion(
+        Question $question,
+        array $answers = null,
+        $canEdit = true
+    ) {
+        $answersDatas = is_null($answers) ? [] : $answers;
```
stefk: It would be better to remove this line and simply use a more precise default parameter:
```php
private function displaySimpleTextQuestion(
    Question $question,
    array $answers = [],
    $canEdit = true
)
```
If it's not possible at least it can be shortened to:
```php
$answersDatas = $answers ?: [];
```

plugin/survey/Entity/Answer/SimpleTextQuestionAnswer.php
```diff
@@ -0,0 +1,71 @@
...
+     * @ORM\Column(name="answer_content", type="string", length=255, nullable=true)
+     */
+    protected $content;
+
+    public function getId()
+    {
+        return $this->id;
+    }
+
+    public function setId($id)
```
stefk: Probably not needed.

stefk:
LGTM

stefk:
@CPASimUSante, the "no-squash" label is reserved to PRs with a clean commit history. Here we have only one significative commit followed by a merge, then multiple duplicate "fix" commit, so a squash is needed IMO. See https://groups.google.com/forum/?hl=fr#!topic/claro-devs/n9Oh-DnBmT8

CPASimUSante:
@stefk sorry, i knew it wasn't the right tag, but when i tested on my computer, i had a problem (acces not granted to open the ressource, though it was) so i wanted to pause the review. But, since on the Docker it was ok, and this kind of problem isn't strictly related to the modifications, it's ok then.
maybe a new tag : no LGTM could be set when you make a PR and afterwards find a bug or a potential bug, before it was merged ? Anyway, it's ok now.

stefk:
@CPASimUSante, you can use the "needs-change" label or write a short comment in that case.

CPASimUSante:
@stefk ok, will do next time. Thanks

plugin/survey/Manager/SurveyManager.php
```diff
@@ -648,6 +660,64 @@ public function getOpenEndedAnswersBySurveyAndQuestionWithoutPager(
         );
     }
 
+    /*******************************************************
+     * Access to SimpleTextQuestionAnswerRepository methods *
+     *******************************************************/
```
Elorfin: Not a real problem but try to avoid this kind of comment blocks. Stick to the usual `//`, `/* */` or PHP doc blocks (`/** */`).

Elorfin:
LGTM.

@CPASimUSante : can you fix the conflicts please ?

Elorfin:
LGTM

PR #1119, opened 2016-09-09
[ResourceManager] fixes non standard style for unpublished resource 

Elorfin:
The property `filter` is not supported in many browser.
It has been replaced by `opacity: 0.4`.

The value can be discussed.

### DIFF
--- a/main/core/Resources/less/layout.less
+++ b/main/core/Resources/less/layout.less
@@ -776,7 +776,7 @@ ul.desktop-parameters-menu
             margin: 5px 0 5px 15%;
             cursor: pointer;
             &.unpublished {
-                opacity: 0.4;
+                opacity: 0.5;
             }
         }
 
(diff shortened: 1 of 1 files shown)


stefk:
Personnaly after some testing I wouldn't go under 0.5 or even 0.6 (remember the resource name and the menu actions should stay readable even if the contrast is low).

Elorfin:
The opacity only impact the icon. Not the name nor the menu.
Is the case in your instance ?

I can set it to a higher value.

stefk:
Nevermind, I've just set it on the whole node. Anyway, 0.5 is still a best fit IMO for icons/thumbnails that have an already light background (yellow, light green, etc.).

Elorfin:
@stefk : done.

stefk:
LGTM

Elorfin:
LGTM

stefk:
LGTM

PR #1128, opened 2016-09-09
[CoreBundle] Allow publication changes on multiple resources

### LINKED ISSUE #982: [CoreBundle]  Ajout de 2 boutons publier / dépublier dans la barre d'outils du gestionnaire de ressources
Dans la barre d'outils du gestionnaire de ressources, il serait intéressant d'jouter 2 boutons qui permettrait de publier ou dépublier une ressource sur un ensemble de ressources sélectionnées.

Actuellement, il faut une par une ouvrir le menu contextuel puis Editer les propriétés puis décocher Publier puis OK

stefk:
Fixes #982.

- ~~I needed symetrical icons for publish/unpublish actions, so I went for [this one](http://fontawesome.io/icon/caret-square-o-left/) and [this one](http://fontawesome.io/icon/caret-square-o-right/), but obviously it isn't great...~~ Finally used eye icons.
- I noticed a difference between permission masks associated with resource properties edition and the permission checks done in the controllers. The last ones were somewhat weaker so I raised them to `ADMINISTRATE` (previously it was `EDIT`).
- The right to see unpublished resources is currently hard-coded in the resource manager/repository, so even if you have the `ADMINISTRATE` permission, you cannot see them unless you are the owner or the workspace manager (ie if you unpublish them they'll disappear for you...). That behaviour is far beyond the scope of this PR so I left it untouched.


stefk:
@Elorfin, I restored the grayscale filter because users were used to see black/white icons for unpublished resources before it gets buggy. It won't work in IE but I guess opacity should be enough there (I haven't tested though...).

kitan1982:
LGTM

ptsavdar:
LGTM

PR #1149, opened 2016-09-13
[PathBundle] fixes save button always disabled after publish

Elorfin:
If you modify a path after publishing it, the save button was never enabled.


stefk:
LGTM

ptsavdar:
LGTM

PR #1153, opened 2016-09-14
[ExoBundle] fixes Hole import

Elorfin:
There was a SQL error when trying to import an Exercise with a Question with holes.


stefk:
LGTM

ptsavdar:
LGTM

PR #1155, opened 2016-09-15
Add issue and pr templates

stefk:
ping @claroline/commiters


ptsavdar:
LGTM

stefk:
@Elorfin, it's backward compatibility breaks. For example, changes in a core api used by plugins.

stefk:
I agree. The versionning strategy is also closely related.

PR #1159, opened 2016-09-15
[PathBundle] opens secondary resource in a new tab

Elorfin:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no

This PR forces new tabs for opening secondary resources.
I know it's a best practice to let the choice but it bothers the users since I have removed it. It's also compliant with the editor which uses new tabs too.


purplefish32:
Code is LGTM , although I am not sure this is a good idea

stefk:
LGTM

PR #1163, opened 2016-09-16
[CoreBundle] replaces webpack img loader by url loader

Elorfin:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no

The image loader only copies the original into the `dist` directory and return an URL to this file.
The url loader includes the image directly into the JS file (base64 encoded based on what I've seen). This is more convenient.


purplefish32:
LGTM

stefk:
LGTM

PR #1169, opened 2016-09-17
[ClaroDashboardBundle] - Minimal version

pitrackster:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Fixed tickets | no

This is the **very first** version of Claroline dashboard tool.
This tool is attended to allow users to create view on differents activities.
This tool uses Claroline logs which are not fully implemented.

For now users can only : 
- Select a workspace
- View the time spent on this workspace 
    - For all users if they are owner of the workspace 
    - Only for them if they subscribe to the workspace


stefk:
@pitrackster, I quickly tested on http://pr-1169-1474115649.travis-build.claroline.net/ but the connection time seems to be stucked on 0.

Also there're two minor problems in the UI:

- the "Vue" button has a different font
- the loading spinner has an odd placement and style

pitrackster:
@stefk yes I saw for the connection time stucked to 0 and the two others UI problems.
I'm trying to resolve those 3 problems. Also need to investigate how logs are written in DB

ptsavdar:
@pitrackster If you have any questions about logs, I would be glad to help if I can. I've done a thing or two with logs and I know more or less how it works.

pitrackster:
@ptsavdar Hi ! Thanks for the proposal ! I think I will need it at a time. For now I'm a bit out of time and will stick to minimal version and use of logs...

pitrackster:
If anyone want to see this very minimal version go here : http://pr-1169-1474384522.travis-build.claroline.net/

Time calculation is based on workspace-enter events. 
It seems that those event are not registered every time... But I might be mistaken...

ptsavdar:
After a few tests I found out that the dashboard in a workspace won't load. Here's the URL http://pr-1169-1474384522.travis-build.claroline.net/claroline/workspaces/158/open/tool/dashboard it's the dashboard in Admin user's personal workspace.

pitrackster:
@ptsavdar Thanks ! This is not supposed to be a workspace tool ! I chnaged this param in plugin config.yml file

pitrackster:
new link for tests : http://pr-1169-1474461596.travis-build.claroline.net/

ptsavdar:
I've re-tested and I didn't encounter any bugs. However I have a question about how exactly the tool tracks time. I'll try to explain myself. For the user 'admin' I created a dashboard for his personal workspace. Then I created a new user, gave the necessary access to this user and did some actions on the workspace. When I went to see the dashboard review about the time spent on workspace, my new user was still 0. Did I miss something?

LaurentGruber:
LGTM

pitrackster:
@ptsavdar for now the time calculation is ... well at least unprecise! 

The process follows this steps : 
- fore ach user concerned by the workspace (only me if I'm not the creator of the workspace, all users that subscribed to the workspace + myself if I'm the owner)
- search for the first enter event on the requested workspace
  - then search for a "out-of-this workspace" event (for now this kind of envent does not exists so I only consider the next "enter-workspace" event for another workspace)
  - if such event exists then make a time diff between the two dates and add it to the total time (only if time is between 5s min and 2 hours max)
    - the 'out-of-workspace' envent date become the new reference date that I'll use to search for the next 'enter-workspace' event , if I find one search for an "out-of-this-workspace" event and so on...
  - if not you've got a 0

It means that if you want to have a time superior to 0 you have to enter the concerned workspace wait for a few minutes and enter another workspace.

But I have seen that the "enter-workspace" events are only written in DB if you open your workspace from "Workspace" (using the drop down list) not "Desktop" (using the top menu bar)

And even then some times the event does not appear in DB...

ptsavdar:
@pitrackster Ok got it. We had a similar case in Spiral Connect, where we had to measure the mean time spent on a workspace. The way we "resolved" the issue was to calculate the time difference between the first and the last action made on the workspace. So, here in our case you could search for the 'workspace-enter' as your start time and for the end time, find the latest action with the same session id made on the workspace (whatever this action might be, blog comment, path step validation etc.) since all the resources are part of the (same) workspace. This way you don't need the 'exited-workspace-event'. You can add it later of course. But imagine the case where someone works simultaneously in two workspaces in different tabs. The entry in the second workspace will trigger the 'exit' event for the first workspace, the user however continues to work in workspace1 in his first tab. What do you think?

pitrackster:
@ptsavdar yes your solution definitly make sense ! I was to focused on workspace-* events to see the session_id field in claro_log table ! this could improve the time calculation a lot ! That's great ! Thanks :-D

ptsavdar:
@pitrackster You're welcome! I had to deal with the same issue a few years ago and did a lot of "experiments" in order to find what it seemed to be the "most optimized" solution. And the feedback I got throughout these years was surprisingly positive! (Surprisingly even for me :stuck_out_tongue: ) So if I can contribute in any way or if you have any questions please let me know.

composer.json
```diff
@@ -137,7 +137,8 @@
         "innova/media-resource-bundle": "*",
         "ujm/exo-bundle": "*",
         "icap/lesson-bundle": "*",
-        "formalibre/office-connect-bundle": "*"
+        "formalibre/office-connect-bundle": "*",
+        "claroline/dashboard-bundle": "*"
```
stefk: This plugin doesn't exist as a standalone package, there's no need to add it to the `replace` section.

main/core/Controller/API/Workspace/WorkspaceController.php
```diff
@@ -93,6 +93,27 @@ public function getUserWorkspacesAction(User $user)
...
+    {
+        $workspaces = $this->workspaceManager->getWorkspacesByUser($user);
+        $result = [];
+        foreach ($workspaces as $workspace) {
+            $exported = $this->workspaceManager->exportWorkspace($workspace);
+            $result[] = $exported;
+        }
+
+        return $result;
+    }
```
stefk: This is a typical `map` operation:

```php
return array_map(function ($workspace) {
    return $this->workspaceManager->exportWorkspace($workspace);
}, $this->workspaceManager->getWorkspacesByUser($user));

```

main/core/Manager/WorkspaceManager.php
```diff
@@ -320,6 +320,31 @@ public function getWorkspacesByUser(User $user)
...
+        'selfRegistration' => $workspace->getSelfRegistration(),
+        'registrationValidation' => $workspace->getRegistrationValidation(),
+        'selfUnregistration' => $workspace->getSelfUnregistration(),
+        'creationDate' => $workspace->getCreationDate(),
+        'isPersonal' => $workspace->isPersonal(),
+        'startDate' => $workspace->getStartDate(),
+        'endDate' => $workspace->getEndDate(),
+        'isAccessDate' => $workspace->getIsAccessDate(),
+        'type' => $workspace->getWorkspaceType(),
+        ];
```
stefk: Spacing/indentation issue here.

main/core/Resources/modules/authentication/Services/UserService.js
```diff
@@ -0,0 +1,28 @@
...
+    this.$http
+            .get(
+                this.UrlService('api_get_connected_user', {})
+            )
+            .success(function onSuccess(response) {
+              deferred.resolve(response)
+            }.bind(this))
+            .error(function onError(response) {
+              deferred.reject(response)
+            })
```
stefk: Same here.

main/core/Resources/modules/users/Service/UserAPIService.js
```diff
@@ -1,28 +1,32 @@
...
+  removeFromCsv(formData) {
+    return this.$http.post(
+            this.UrlService('api_csv_remove_user'),
+            formData, {
+              transformRequest: angular.identity,
+              headers: {
+                'Content-Type': undefined
+              }
             }
         )
```
stefk: Same here:

```js
return this.$http.post(
  this.UrlService('api_csv_remove_user'),
  formData, 
  {
    transformRequest: angular.identity,
    headers: {
       'Content-Type': undefined
    }
  }
)
```

main/core/Resources/modules/users/Service/UserAPIService.js
```diff
@@ -1,28 +1,32 @@
...
+    return this.$http.post(
+            this.UrlService('api_csv_import_facets'),
+            formData, {
+              transformRequest: angular.identity,
+              headers: {
+                'Content-Type': undefined
+              }
             }
         )
-    }
```
stefk: Idem.

main/core/Resources/modules/workspace/Services/WorkspaceService.js
```diff
@@ -0,0 +1,27 @@
+
+
```
stefk: Superfluous empty lines.

main/core/Resources/modules/workspace/Services/WorkspaceService.js
```diff
@@ -0,0 +1,27 @@
...
+class WorkspaceService{
```
stefk: Could be `export default class WorkspaceService {` (without the final `export`).

plugin/dashboard/ClarolineDashboardBundle.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Claroline\DashboardBundle;
+
+use Claroline\CoreBundle\Library\PluginBundle;
+use Claroline\KernelBundle\Bundle\ConfigurationBuilder;
+
+/**
+ * Bundle class.
+ * Uncomment if necessary.
```
stefk: The second sentence makes no sense and should be removed.

plugin/dashboard/Manager/DashboardManager.php
```diff
@@ -0,0 +1,209 @@
...
+    {
+        return $this->em->getRepository('ClarolineCoreBundle:Log\Log');
+    }
+
+    /**
+     * get all dashboards for a given user.
+     */
+    public function getAll(User $user)
+    {
+        $result = [];
```
stefk: Also a `map`.

plugin/dashboard/Controller/Api/DashboardsController.php
```diff
@@ -0,0 +1,158 @@
...
+    /**
+     * @EXT\Route("/dashboards/count", name="get_nb_dashboards")
+     * @EXT\ParamConverter("user", converter="current_user", options={"allowAnonymous"=false})
+     * @EXT\Method("GET")
+     */
+    public function getNbDashboard(User $user)
+    {
+        $dashboards = $this->dashboardManager->getAll($user);
+
+        return new JsonResponse(count($dashboards));
```
stefk: It would be way better to have a proper `COUNT` query. Dashboard records are fetched and exported for nothing.

plugin/dashboard/Controller/Api/DashboardsController.php
```diff
@@ -0,0 +1,158 @@
...
+     * @EXT\Route("/dashboards/{dashboardId}/times", name="get_dashboard_spent_times")
+     * @EXT\ParamConverter("user", converter="current_user", options={"allowAnonymous"=false})
+     * @EXT\ParamConverter("dashboard", class="ClarolineDashboardBundle:Dashboard", options={"mapping": {"dashboardId": "id"}})
+     * @EXT\Method("GET")
+     */
+    public function getDashboardWorkspaceSpentTimes(User $user, Dashboard $dashboard)
+    {
+        $all = $user->getId() === $dashboard->getWorkspace()->getCreator()->getId();
+
+        $datas = $this->dashboardManager->getDashboardWorkspaceSpentTimes($dashboard->getWorkspace(), $user, $all);
```
stefk: Minor: `data` is always singular.

plugin/dashboard/Controller/Api/DashboardsController.php
```diff
@@ -0,0 +1,158 @@
...
+    /**
+     * @EXT\Route("/dashboards/preview/{workspaceId}/times", name="get_dashboard_spent_times_by_workspace")
+     * @EXT\ParamConverter("user", converter="current_user", options={"allowAnonymous"=false})
+     * @EXT\ParamConverter("workspace", class="ClarolineCoreBundle:Workspace\Workspace", options={"mapping": {"workspaceId": "id"}})
+     * @EXT\Method("GET")
+     */
+    public function getDashboardWorkspaceSpentTimesByWorkspace(User $user, Workspace $workspace)
+    {
+        $all = $user->getId() === $workspace->getCreator()->getId();
+        $datas = $this->dashboardManager->getDashboardWorkspaceSpentTimes($workspace, $user, $all);
```
stefk: Same.

plugin/dashboard/Controller/Api/DashboardsController.php
```diff
@@ -0,0 +1,158 @@
...
+     * @EXT\Route("/dashboards/{dashboardId}/delete", name="delete_dashboard")
+     * @EXT\ParamConverter("user", converter="current_user", options={"allowAnonymous"=false})
+     * @EXT\ParamConverter("dashboard", class="ClarolineDashboardBundle:Dashboard", options={"mapping": {"dashboardId": "id"}})
+     * @EXT\Method("DELETE")
+     */
+    public function deleteDashboard(User $user, Dashboard $dashboard)
+    {
+        $this->dashboardManager->delete($dashboard);
+        // return all existing dashboards
+        return $this->getAll($user);
```
stefk: Is this extra request really necessary?

plugin/dashboard/Controller/DashboardController.php
```diff
@@ -0,0 +1,12 @@
+<?php
```
stefk: Empty class, should be removed.

plugin/dashboard/Manager/DashboardManager.php
```diff
@@ -0,0 +1,209 @@
...
+        if ($all) {
+            // get all users involved in the workspace
+            $ids = $this->getWorkspaceUsersIds($workspace);
+        } else {
+            // only the current user
+            $ids[] = $user->getId();
+        }
+
+        // for each user (ie user ids) -> get first 'workspace-enter' event for the given workspace
+        foreach ($ids as $uid) {
```
stefk: It has to be a better way to fetch and compute data. Making sql queries in a loop is a bad idea. Here you're making two of them plus a nested iteration in `computeTimeForUserAndWorkspace`. That's terrible for performance.

plugin/dashboard/Manager/DashboardManager.php
```diff
@@ -0,0 +1,209 @@
...
+            $userSqlSelect = 'SELECT first_name, last_name FROM claro_user WHERE id = '.$uid;
```
stefk: This one could probably be removed if `getWorkspaceUsersIds` was replaced by something like `getWorkspaceUsers`.

plugin/dashboard/Manager/DashboardManager.php
```diff
@@ -0,0 +1,209 @@
...
+                $time += $seconds;
+            }
+            // get next "enter the requested workspace enter event"
+            $sql = 'SELECT date_log FROM claro_log WHERE action LIKE "workspace-enter" AND doer_id = '.$userId.' AND date_log > "'.$action['date_log'].'" AND workspace_id = '.$workspaceId.' ORDER BY date_log ASC LIMIT 1';
+            $stmt = $this->em->getConnection()->prepare($sql);
+            $stmt->execute();
+            $nextEnterEvent = $stmt->fetch();
+            // if there is an "enter-workspace" action after the current one recall the method
+            if ($nextEnterEvent['date_log']) {
+                return $this->computeTimeForUserAndWorkspace($nextEnterEvent['date_log'], $userId, $workspaceId, $time);
```
stefk: It's even recursive!

plugin/dashboard/Resources/less/loading-bar.less
```diff
@@ -0,0 +1,119 @@
...
+        border: 0;
+
+        &:after {
+            content: '';
+            display: block;
+            font-size: 10px;
+            width: 1em;
+            height: 1em;
+            margin-top: -0.5em;
+            -webkit-animation: spinner 1500ms infinite linear;
```
stefk: We use [autoprefixer](https://github.com/postcss/autoprefixer), so vendor prefixed aren't needed.

plugin/dashboard/Resources/modules/dashboard/Controllers/DetailsCtrl.js
```diff
@@ -0,0 +1,14 @@
+
```
stefk: Emtpy class, should be removed.

plugin/dashboard/Resources/modules/dashboards/Filters/SecondsToHmsFilter.js
```diff
@@ -0,0 +1,31 @@
...
+      if (hours < 10) {
+        hours = '0' + hours
+      }
+      if (minutes < 10) {
+        minutes = '0' + minutes
+      }
+      if (seconds < 10) {
+        seconds = '0' + seconds
+      }
+       // var time = hours + ':' + minutes + ':' + seconds + ':' + ms;
```
stefk: Commented code, should be removed.

plugin/dashboard/Resources/translations/dashboard.en.yml
```diff
@@ -0,0 +1,52 @@
+# A
```
stefk: A lot of translations here are already in the `platform` translation domain. They should be used instead.

composer.json
pitrackster: done

main/core/Controller/API/Workspace/WorkspaceController.php
pitrackster: done

main/core/Manager/WorkspaceManager.php
pitrackster: done

main/core/Resources/modules/authentication/Services/UserService.js
pitrackster: done

main/core/Resources/modules/users/Service/UserAPIService.js
pitrackster: done

main/core/Resources/modules/users/Service/UserAPIService.js
pitrackster: done

main/core/Resources/modules/workspace/Services/WorkspaceService.js
pitrackster: done

main/core/Resources/modules/workspace/Services/WorkspaceService.js
pitrackster: done

plugin/dashboard/ClarolineDashboardBundle.php
pitrackster: done

plugin/dashboard/Manager/DashboardManager.php
pitrackster: done

plugin/dashboard/Controller/Api/DashboardsController.php
pitrackster: done

plugin/dashboard/Controller/Api/DashboardsController.php
pitrackster: done

plugin/dashboard/Controller/Api/DashboardsController.php
pitrackster: done

plugin/dashboard/Controller/DashboardController.php
pitrackster: done

plugin/dashboard/Resources/less/loading-bar.less
pitrackster: done

plugin/dashboard/Resources/modules/dashboard/Controllers/DetailsCtrl.js
pitrackster: done

plugin/dashboard/Resources/modules/dashboards/Filters/SecondsToHmsFilter.js
pitrackster: done

plugin/dashboard/Resources/translations/dashboard.en.yml
pitrackster: done

plugin/dashboard/Controller/Api/DashboardsController.php
```diff
@@ -139,8 +139,6 @@ public function getDashboardWorkspaceSpentTimesByWorkspace(User $user, Workspace
     public function deleteDashboard(User $user, Dashboard $dashboard)
     {
         $this->dashboardManager->delete($dashboard);
-        // return all existing dashboards
-        return $this->getAll($user);
```
stefk: You still need to return a response here, otherwise an exception will be thrown. It could be a simple 204 (no content) json response.

pitrackster: done

LaurentGruber:
un ticket a été fait avec les changements encore nécessaire : https://github.com/claroline/Distribution/issues/1271. On va le merger comme ça pour le moment afin d'inclure la preview dans la 16.09

LaurentGruber:
LGTM

stefk:
tmp LGTM (#1271)

PR #1170, opened 2016-09-19
[ExoBundle] Fix score computation when no solutions are given

stefk:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no

Some parts of the code currently assume that a `solution` property is always present on `question` objects, leading to "xyz is not defined" errors and eventually breaking whole exercises.


Elorfin:
LGTM

kitan1982:
LGTM

PR #1171, opened 2016-09-19
[FormulaBundle] Exclude non element nodes from inspection

stefk:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no

It seems that the `NodeChange` event triggered at initialization time is also applied to comment nodes, leading to an "undefined" error on `className` access when the original html value contains comments.


ptsavdar:
LGTM

kitan1982:
LGTM

stefk:
LGTM

PR #1174, opened 2016-09-19
[PortfolioBundle] Portfolio improvements

### LINKED ISSUE #991: Portfolio : improvements
- in the Experience tab, the field "website link" is mandatory : it should not, IMO
- When i click on a portfolio in the Portfolio tab (the portfolio link or the update button), i can't go back to the edition mode 
- in the portfolio view, the way of adding/ ordering items is tedious : if i have set 3 competencies, 3 experiences, 3 training i have to add the 9 items. It could have been 3 : a block for each (competency, Xp and traing) with all the items; or if you want to choose which element to add, at least allow a checkbox instead a radio.

To be clear, only the first point is problematic in term of funcionnality.
The two other are just ergonomy improvement.

ptsavdar:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Fixed tickets | fixes #991 

- Rendered website url in experience widget optional
- Added back button to view mode (so as user can return to edit mode)
- User can now insert multiple widgets of the same type at once

### DIFF
--- a/plugin/portfolio/Resources/modules/checklist-model.js
+++ b/plugin/portfolio/Resources/modules/checklist-model.js
@@ -1,0 +1,5 @@
+/**
+ * Created by panos on 9/16/16.
+ */
+import {} from 'checklist-model';
+import _ from 'lodash';
--- a/plugin/portfolio/Resources/modules/checklist-model.js
+++ b/plugin/portfolio/Resources/modules/checklist-model.js
@@ -1,5 +1,5 @@
 /**
  * Created by panos on 9/16/16.
  */
-import {} from 'checklist-model';
-import _ from 'lodash';
+import {} from 'checklist-model'
+import {} from 'lodash'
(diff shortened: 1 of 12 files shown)


plugin/portfolio/Resources/modules/checklist-model.js
```diff
@@ -0,0 +1,5 @@
+/**
+ * Created by panos on 9/16/16.
+ */
+import {} from 'checklist-model'
+import {} from 'lodash'
```
stefk: Can you add lodash to package.json?

ptsavdar: I think lodash is already imported through bower, isn't it? 
Should I import it 2 times or remove the bower dependency?

stefk: Sorry, I forgot that. We can leave it as is.

stefk:
LGTM

Elorfin:
LGTM

PR #1200, opened 2016-09-21
[CoreBundle] - directive allowing to disable a link. closes #1199

### LINKED ISSUE #1199: [ExoBundle] - Player - Try Exercise button
| Q             | A
| ------------- | ---
| Bug report?   | yes
| Version?      | master

When trying an exercise the [Test exercise] link is (style) disabled but is still active.
If we navigate through steps and then click the link we go back to first step...

pitrackster:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| Fixed tickets | #1199

Core directive that allow to add a "disable" attribute on a link

### DIFF
--- a/plugin/exo/Resources/modules/exercise/module.js
+++ b/plugin/exo/Resources/modules/exercise/module.js
@@ -10,7 +10,7 @@ import 'angular-ui-tinymce'
 import 'angular-ui-translation/angular-translation'
 import '#/main/core/modal/module'
 import '#/main/core/fos-js-router/module'
-import '#/main/core/aDisabled/module'
+import '#/main/core/anchor/module'
 
 import './../common/module'
 import './../feedback/module'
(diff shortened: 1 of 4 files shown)


plugin/exo/Resources/modules/exercise/module.js
```diff
@@ -10,6 +10,7 @@ import 'angular-ui-tinymce'
 import 'angular-ui-translation/angular-translation'
 import '#/main/core/modal/module'
 import '#/main/core/fos-js-router/module'
+import '#/main/core/aDisabled/module'
```
Elorfin: Use same convention as other modules : `a-disabled`.
Also, i would have created a `link` module which could contain more than this sole directive.

plugin/exo/Resources/modules/exercise/module.js
```diff
@@ -32,6 +33,7 @@ angular
     'ui.tinymce',
     'ui.modal',
     'mgcrea.ngStrap.datepicker',
+    'LinkDisabled',
```
Elorfin: Other core modules uses lowercase.

LaurentGruber:
LGTM

PR #1225, opened 2016-09-26
[ExoBundle] Solve several problems on bind questions

pitrackster:
| Q             | A
| ------------- | ---
| Bug fix?      | yes

Solve the following issues : 

- Delete connection was possible through Drag&Drop but the answer was kept
  - We can't remove a connection via Drag&Drop anymore
-  In the instant feedback user was allowed to remove a connection via Drag&Drop
 - By removing the abilitty to remove a connection via Drag&Drop this is not possible anymore
- After the MacthQuestion directive ```destroy``` method, some jsPlumb events where still alive which was the cause a lot of problems. 
  - The use of ```jsPlumb.reset()``` solve this by removing every existing events

### DIFF
--- a/plugin/exo/Resources/modules/question/Directives/Type/MatchQuestionDirective.js
+++ b/plugin/exo/Resources/modules/question/Directives/Type/MatchQuestionDirective.js
@@ -34,12 +34,6 @@ function MatchQuestionDirective(FeedbackService, $timeout, $window, MatchQuestio
               return controller.handleBeforeDrop(info)
             })
 
-            /*jsPlumb.bind('beforeDetach', function(connection){
-              // According to the doc Returning false - or nothing - from this callback will cause the detach to be abandoned, and the Connection will be reinstated or left on its current target.
-              // But if we return false User will not be allowed to remove connection by clicking on it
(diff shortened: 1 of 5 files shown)


plugin/exo/Resources/modules/question/Directives/Type/MatchQuestionDirective.js
```diff
@@ -34,6 +34,16 @@ function MatchQuestionDirective(FeedbackService, $timeout, $window, MatchQuestio
               return controller.handleBeforeDrop(info)
             })
 
+            /*jsPlumb.bind('beforeDetach', function(connection){
+              // According to the doc Returning false - or nothing - from this callback will cause the detach to be abandoned, and the Connection will be reinstated or left on its current target.
+              // But if we return false User will not be allowed to remove connection by clicking on it
+              // If we return nothing this will be possible
+            })*/
```
Elorfin: No commented code.

Elorfin:
LGTM

LaurentGruber:
LGTM

Elorfin:
LGTM

Issue #1208, opened 2016-09-22
[PathBundle] Adaptive content as a primary resource : display issue


Pauooo:
| Q             | A
| ------------- | ---
| Bug report?   | yes
| Version?      | master

When the primary resource of a step is an adaptive content (for exemple a SCORM), the iframe height is oversized. It makes it impossible to display properly the SCORM in a step.
http://claroline-master.travis-build.claroline.net/innova_path/player/6#/8

LaurentGruber:
voir pr : https://github.com/claroline/Distribution/pull/878

PR #1267, opened 2016-09-29
[CoreBundle] Iframe resizing.

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Fixed tickets |  #1208

#1208 goes into infinite loop because there is always a 150px difference between the old and the new height (I don't know why, maybe the scorm automatically does something similar) so I added a 150px interval and it fixes it.

I ran into some similar issues before but I only had to add a 5px difference (so it's totally ok). It's really huge here.

### DIFF
--- a/main/core/Resources/views/Workspace/layout.html.twig
+++ b/main/core/Resources/views/Workspace/layout.html.twig
@@ -116,12 +116,16 @@
             var getIframeWindowHeight = function() {
                 return $("iframe#embeddedActivity").contents().find('body').first().height();
             }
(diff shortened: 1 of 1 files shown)


kitan1982:
LGTM

LaurentGruber:
à vérifier avec d'autres scorm

LaurentGruber:
close for https://github.com/claroline/Distribution/pull/1370

PR #1274, opened 2016-09-29
User import speed improvement

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no

The check takes a long time so I disabled it for large user import (the RoleManager::validateRoleInsert takes a long time to execute and should not be required for the csv import as the validation must be able to handle it).

The check performance and validation prior to the import are separate issues IMO.


ngodfraind:
These 2 commits increase the performances by a factor of 5 to 10 imo. Probably even more. It's really a lot faster.

kitan1982:
LGTM

LaurentGruber:
LGTM

PR #1311, opened 2016-10-04
[CoreBundle] Fixing tags in workspace list.

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no

Checking if the array have the correct keys because it appears to not always be the case.


kitan1982:
LGTM

Elorfin:
LGTM

PR #1319, opened 2016-10-05
[PathBundle] fixes #1273

### LINKED ISSUE #1273: [PathBundle] Sauvegarde impossible si le parcours n'est pas publié au départ
| Q             | A
| ------------- | ---
| Bug report?   | yes
| Version?      | master

Créer un parcours en décochant Publier la ressource
Editer le parcours
Impossible de Sauvegarder
image

Elorfin:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Fixed tickets | #1273


ngodfraind:
I understand what you're doing and it's sort of fine as a bugfix but do you really need the id ? (is it possible to handle everything with the guid of the path instead ?)

I'll have to do some changes at the workspace import level because I ran into similar problems and it's pretty annoying overall so if you have an other way to do it you should use it instead.

Otherwise LGTM

Elorfin:
I need the ID to generate routes in Angular (for save / publish).
My problem is there is no GUID on path for now. That's the quickest wsay I've found to make it work.

But yes, when I will have GUIDs on paths, this will be replaced by it.

kitan1982:
LGTM

Issue #647, opened 2016-06-15
Authentification CAS


david42:
CAS not works with "claro_security_login" : 
https://github.com/claroline/Distribution/blob/master/main/core/Controller/LayoutController.php#L197

ngodfraind:
#1345 should have fixed it.

PR #1345, opened 2016-10-07
[ClarolineCoreBundle] Configurable login target url

### LINKED ISSUE #1339: [CoreBundle] Paramétrage du bouton login
| Q             | A
| ------------- | ---
| Bug report?   | no
| Version?      |

Avec O365, on a lié le bouton login directement sur l'authentification o365 mais à chaque mise à jour c'est écrasé.

Ce serait bien d'avoir un paramètre de config de la plateforme du login target

Modification à faire actuellement
Fichier : main/core/Controller/LayoutController.php
Ligne 197: ("$loginTarget = $this->router...")
  -> remplacer 'claro_security_login' par 'claro_o365_get_token'

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Fixed tickets | #1339 

This is very specific so there is no GUI for that one (most ppl will not understand what it is for).
Fixes #1339

### DIFF
--- a/main/core/Controller/Administration/ParametersController.php
+++ b/main/core/Controller/Administration/ParametersController.php
@@ -221,7 +221,6 @@ class ParametersController extends Controller
                             'registration_mail_validation' => $form['registrationMailValidation']->getData(),
                             'is_pdf_export_active' => $form['isPdfExportActive']->getData(),
                             'ssl_enabled' => $form['sslEnabled']->getData(),
-                            'login_target_route' => $form['loginTargetRoute']->getData(),
                         ]
                     );
 
--- a/main/core/Controller/Administration/ParametersController.php
+++ b/main/core/Controller/Administration/ParametersController.php
@@ -221,6 +221,7 @@ class ParametersController extends Controller
                             'registration_mail_validation' => $form['registrationMailValidation']->getData(),
                             'is_pdf_export_active' => $form['isPdfExportActive']->getData(),
                             'ssl_enabled' => $form['sslEnabled']->getData(),
+                            'login_target_route' => $form['loginTargetRoute']->getData(),
                         ]
                     );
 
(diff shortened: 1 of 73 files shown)


ngodfraind:
@ptsavdar 
Why does it works like this ? I can't remember.
This is a change you introduced here: https://github.com/claroline/Distribution/commit/a4cba90d7cf0c5b01addb1c9ec96ec5875228f7f

main/core/Controller/Administration/ParametersController.php
```diff
@@ -221,6 +221,7 @@ public function generalFormAction(Request $request)
                             'registration_mail_validation' => $form['registrationMailValidation']->getData(),
                             'is_pdf_export_active' => $form['isPdfExportActive']->getData(),
                             'ssl_enabled' => $form['sslEnabled']->getData(),
+                            'login_target_route' => $form['loginTargetRoute']->getData(),
```
stefk: This should probably be removed.

ngodfraind: yep, I added it in the form but removed it later as it wasn't a good idea.

ngodfraind:
@david42 
related to #647

LaurentGruber:
LGTM

LaurentGruber:
@ptsavdar cela ne cassera pas la fonctionnalité de choix de redirection après login que tu as ajouté ?

ptsavdar:
@ngodfraind : The redirection after login process works as follows: When a user visits a page anonymously, and at a certain point he decides to login in order to interact using his account, after login he is redirected to the last page he was viewing before login.

@ngodfraind, @LaurentGruber There is a list with excluded routes in [AuthenticationSuccessListener](https://github.com/claroline/Distribution/blob/master/main/core/Listener/AuthenticationSuccessListener.php#L288), this route/url needs to be added to the list of excluded routes and not taken into account as a "last valid visited URL".

purplefish32:
LGTM

PR #1365, opened 2016-10-10
[CoreBundle] Allows the router to generate absolute urls with params …

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no

…from platforms_options.yml.


ngodfraind:
@stefk 

See the ClaroRouter class.

It's pretty hacky. There should have a better way but it works like this.
The main issue I have with it is that I have to rebuild the context every time the generate function is fired (for some reason I don't know yet, the constructor is enough for the CLI only).

ptsavdar:
LGTM

LaurentGruber:
LGTM

LaurentGruber:
LGTM

PR #1366, opened 2016-10-10
[exo-v2] updates and fixes api

Elorfin:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no

- Changes the Exercise API "export" endpoint to use the new API (not used by the player and I needed it to test with a REST client).
- Fixes various bugs in update.
- Adds GUID on questions and uses it in the API.
- Adds more tests
- Finishes implementation for Open / Cloze / Words (need to be tested)
- Makes Question::title optional

In the validation of the Question, I've made the property `score` required.
This is a required info for now, but when we will have implemented exercise without scores, it will become optional. That's why I've not made it directly in the schema (I would have need to update all example files).

This needs **migrations and additional installer to be run**.

ping @pitrackster, @stefk

### DIFF
--- a/plugin/exo/Entity/Question.php
+++ b/plugin/exo/Entity/Question.php
@@ -10,11 +10,14 @@ use Symfony\Component\Validator\Constraints as Assert;
 /**
  * @ORM\Entity(repositoryClass="UJM\ExoBundle\Repository\QuestionRepository")
  * @ORM\Table(name="ujm_question")
+ * @ORM\EntityListeners({"\UJM\ExoBundle\Listener\Entity\QuestionListener"})
  * @ORM\HasLifecycleCallbacks()
  */
 class Question
 {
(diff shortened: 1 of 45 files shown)


Elorfin:
For now, the API is not able to distinguish Words question "one word" and "short".
I use "short" by default but we need to find a format to add it to the Words schema.

plugin/exo/Entity/Question.php
```diff
@@ -10,63 +10,89 @@
...
      */
     private $id;
 
     /**
-     * @ORM\Column
+     * @var string
+     *
+     * @ORM\Column()
+     */
+    protected $guid;
```
stefk: IMO it should be called `uuid` and have an explicit size of 36 characters.

PR #1384, opened 2016-10-11
[exo-v2] adds UUID and cleans Additional Installer

Elorfin:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no

- Adds UUIDs on Exercise and Steps and uses it.
- Generate UUIDs from mysql in updater
- Change updater conditions

@pitrackster @stefk

### DIFF
--- a/plugin/exo/Entity/Exercise.php
+++ b/plugin/exo/Entity/Exercise.php
@@ -19,6 +19,13 @@ class Exercise extends AbstractResource
     const TYPE_FORMATIVE = '3';
 
     /**
+     * @var string
+     *
+     * @ORM\Column("uuid", type="string", length=36)
+     */
+    private $uuid;
+
+    /**
      * @ORM\Column(name="description", type="text", nullable=true)
      */
     private $description = '';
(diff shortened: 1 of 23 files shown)


plugin/exo/Entity/Exercise.php
```diff
@@ -19,6 +19,13 @@ class Exercise extends AbstractResource
     const TYPE_FORMATIVE = '3';
 
     /**
+     * @var string
+     *
+     * @ORM\Column("uuid", type="string", length=36)
+     */
+    protected $uuid;
```
stefk: Why not private?

Elorfin: Just an habit. I always put my Entities properties in `protected`. I've changed it to set it to `private`.

stefk:
@Elorfin, I just realized unique constraints/indexes are needed on uuid columns.

Elorfin:
@stefk You're right. I will add them in my opened PR for the exo v2.

PR #1391, opened 2016-10-12
[exo-v2] updates API

Elorfin:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no

Follows modifications of https://github.com/json-quiz/json-quiz/pull/35.


plugin/exo/Serializer/ExerciseSerializer.php
```diff
@@ -155,9 +155,19 @@ private function serializeParameters(Exercise $exercise)
                 break;
         }
 
-        $parameters->random = $exercise->getShuffle();
+        if ($exercise->getShuffle()) {
+            $parameters->randomOrder = $exercise->getKeepSteps() ? 'once' : 'always';
```
stefk: Ideally we should use constants here: `SHUFFLE_ONCE`, `SHUFFLE_ALWAYS`, `PICK_ONCE`, `PICK_ALWAYS`, etc. (these constants could be reused in various locations).

Elorfin:
I've added the unique constraint on UUIDs according to the comment on #1384.

PR #1394, opened 2016-10-12
Path save/publish fix

CPASimUSante:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | maybe (see below)
| Fixed tickets | https://github.com/claroline/Distribution/issues/1219

**Warning/spoiler alert : not a ideal solution.**

This PR (poorly) fixes the problem of being unable to exit the path edit after publishing.
The problem found lies in the fact that in a $scope.watch there is a test that stays always true after the path init, hence always give the same variable value

```
      if (!empty && updated) {
        // Initialization is already done, so mark path as unsaved for each modification
        this.unsaved = true
      }
```
empty is always true because the history is never cleared.
I cleared this history when publishing. This is a kind of functionnal "regression".

Another side effect is that if you create a path, steps, conditions, save and publish without exiting it goes ok, but if you want to change path header stuff (title, description), the change is not taken into account : no Save button enable. But if you change something else (!)  something in the steps, .... or reload, everything will work again.

This fix also gets rid of the "annoying" popup when exiting. This is because of a problem between angular/tinymce.

I used this code from the fix in https://github.com/claroline/Distribution/issues/1043
=> https://www.tinymce.com/docs/plugins/autosave/#autosave_ask_before_unload


Elorfin:
LGTM.

In fact we must clear history at publication. 

In the history, the ids of the generated entities are not known (steps, conditions, activities, etc.). So if the user use the history and republish a new set of entities is generated (and the old one will remain in DB).

ngodfraind:
LGTM

ngodfraind:
LGTM

PR #1408, opened 2016-10-13
[CoreBundle] Modernizr fixes #1399

### LINKED ISSUE #1399: Add a warning in unsupported browsers
| Q             | A
| ------------- | ---
| Bug report?   | no
| Version?      | master

In several places, we rely on rather advanced browser features without performing any check. The resulting bugs can be a huge source of frustration for users. Possible solution:

1. include [modernizr](https://modernizr.com) in every page
2. configure it to detect Mutation Observer support (that should cover our [browser range](http://caniuse.com/#feat=mutationobserver)) and let it add a corresponding class in the `html` tag
3. add a big/obvious/red header in the page layout with a descriptive and helpful message for non supported browsers
4. display that header conditionnaly via css depending on the class set by modernizr

Notes:
- conditional comments like [this one](https://github.com/claroline/Distribution/blob/master/main/core/Resources/views/base.html.twig#L41) are useless in this case and should be removed
- if size is an issue, modernizr can be stripped down to perform only the checks we need (but then we won't be able to detect additional/optional features)
- this should obviously be tested in various browsers

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Fixed tickets | #1399

What do we add in the alert as an error message ?


package.json
```diff
@@ -28,6 +28,8 @@
     "karma-webpack": "^1.7.0",
     "less": "^2.5.3",
     "mocha": "^2.5.3",
+    "modernizr": "https://registry.npmjs.org/modernizr/-/modernizr-3.3.1.tgz",
```
stefk: ?

ngodfraind: copy pasted from what was generated

stefk:
Why didn't you use the css approach?

package.json
stefk: doesn't seem right...

ngodfraind:
A css class is added automatically and we can define its content ourself, but I have no idea how I could add a div like that with css only, so I used js to do this.

package.json
ngodfraind: `npm install --save-dev modernizr modernizr-loader`

stefk:
> I have no idea how I could add a div like that with css only

Just add a div in base.html.twig. Put a class ".outdated-browser-warning" on it. Then in the main css, add two rules like that: 

```css
.outdated-browser-warning {
  display: none;
  /* some styles here */
}

.no-mutationobserver .outdated-browser-warning {
  display: box;
}
```

ngodfraind:
Can we remove  `import modernizr from 'modernizr'` and the file ?

stefk:
Proposal:

EN:

> It appears that you're using an unsupported web browser. To enjoy a pleasant and bug-free experience, please use a recent version of a major browser (http://browsehappy.com/). Internet Explorer is supported from version 11.

FR:

> Vous utilisez un navigateur non pris en charge. Pour une éviter tout fonctionnement indésirable, veuillez utiliser une version récente d'un navigateur majeur (http://browsehappy.com/). Internet Explorer est supporté à partir de la version 11.

stefk:
Can you provide a screenshot?

package.json
stefk: You can safely replace it by:
```json
"modernizr": "^3.3.1"
```
That's the (expected) output I get from that command.

ngodfraind:
The behaviour is different:
dev / watch is working
prod /webpack isn't

stefk:
If I remember correctly, that's what happen when `commons.js` isn't included before the bundle.

ngodfraind:
It is related to CommonsChunkPlugin

stefk:
You probably need this : https://webpack.github.io/docs/list-of-plugins.html#2-explicit-vendor-chunk
(one single chunk for modernizr)

ngodfraind:
It looks like it's working when I add modernizr at the end of the scripts.

ngodfraind:
screenshot from 2016-10-14 09-55-56

stefk:
@ngodfraind I'd add a fa-warning icon before the text and make the browsehappy url clickable, otherwise LGTM.

ngodfraind:
LGTM

PR #1416, opened 2016-10-14
[VideoPlayerBundle] The generated html is now easily editable.

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Fixed tickets | #1117

### DIFF
--- a/plugin/video-player/Resources/views/player.html.twig
+++ b/plugin/video-player/Resources/views/player.html.twig
@@ -1,22 +1,25 @@
-<div class="container">
-    <div class="span12">
-        <div>
-            <video
-                controls
-                class="video-js vjs-big-play-centered vjs-default-skin vjs-16-9"
-                data-setup='{}'
-                >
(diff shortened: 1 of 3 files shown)


plugin/video-player/Resources/views/player.html.twig
```diff
@@ -1,22 +1,25 @@
...
-                          {% if (track.isDefault()) %} default {% endif %}
-                      >
-                 {% endfor %}
-            </video>
-        </div>
-    </div>
+
+<div style="display:none;">
+    {{ 'html_explain'|trans({}, 'videojs') }}
+</div>
```
Elorfin: IMHO, it seems a little heavy to add this in the markup every time we add a video. 
We will store it in DB many times.

ngodfraind: I think it's negligible and it allows users to know what do to do position/resize their video (this is not very intuitive otherwise).

LaurentGruber:
LGTM

ngodfraind:
LGTM

Issue #1028, opened 2016-08-31
[ExoBundle]


ngodfraind:
I just crashed everything for the 100 times once again because I forgot it.
This is just a reminder for me.

LaurentGruber:
@ngodfraind tu sais pourquoi tu avais fais ce ticket ?

PR #1427, opened 2016-10-17
[CoreBundle] CLI doctrine migration handling.

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Fixed tickets | #1028

This simple CLI tool allows us to add entries in doctrine table when they're missing easily without having to go into mysql/phpmyadmin and try to find what we need (because these tables are hard to read).

There is also a manager because we used to do that manually in updaters and we could get it wrong.

`php app/console claroline:migration:check clarolinecorebundle 20150428152724`

The previous line will add the entry in the databse if it doesn't exists.

### DIFF
--- a/main/core/Command/DatabaseIntegrity/MigrationsCheckerCommand.php
+++ b/main/core/Command/DatabaseIntegrity/MigrationsCheckerCommand.php
@@ -1,0 +1,48 @@
+<?php
+
+/*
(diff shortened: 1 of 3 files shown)


LaurentGruber:
LGTM

ptsavdar:
LGTM

PR #1442, opened 2016-10-19
[CoreBundle] Check if importer exists (Fixes #1433)

### LINKED ISSUE #1433: [CoreBundle] Bug Import of a WS
| Q             | A
| ------------- | ---
| Bug report?   | yes
| Version?      | master

When I want to create a WS by import I have this error : 
Uncaught PHP Symfony\Component\Debug\Exception\FatalThrowableError Exception: "Fatal error: Call to a member function setWorkspace() on null" at /var/www/html/claroline/vendor/claroline/distribution/main/core/Manager/TransferManager.php line 259

But create the WS adding an "activity_folder" in the resources

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Fixed tickets | #1433

There are no activity importer so we have to check the condition.
We might add it later (at least we check it exists now because there isn't an importer for each resource type).

### DIFF
--- a/main/core/Manager/TransferManager.php
+++ b/main/core/Manager/TransferManager.php
@@ -256,11 +256,14 @@ class TransferManager
 
         foreach ($data['tools'] as $tool) {
             $importer = $this->getImporterByName($tool['tool']['type']);
-            $importer->setWorkspace($workspace);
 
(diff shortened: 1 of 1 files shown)


kitan1982:
LGTM

LaurentGruber:
LGTM

PR #1445, opened 2016-10-19
[CoreBundle] CLI tool for emptying workspace roles.

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Fixed tickets | #1438 

This is the CLI part of #1438

### DIFF
--- a/main/core/Entity/Role.php
+++ b/main/core/Entity/Role.php
@@ -280,6 +280,16 @@ class Role implements RoleInterface
         }
     }
 
+    public function removeUser(User $user)
+    {
+        $user->getRoles()->removeElement($this);
+    }
+
+    public function initUsers()
+    {
+        $this->users = new ArrayCollection();
+    }
+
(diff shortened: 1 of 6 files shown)


LaurentGruber:
LGTM

main/core/Entity/Role.php
```diff
@@ -280,6 +280,16 @@ public function addUser($user)
         }
     }
 
+    public function removeUser(User $user)
+    {
+        $user->getRoles()->removeElement($this);
+    }
```
Elorfin: This seems a little weird. Why not just use/declare `$user->removeRole(Role $role)` ? Also I don't see where you use this new method.

ngodfraind: You're right, it should be remvoed.

Elorfin:
LGTM

Issue #1065, opened 2016-09-05
[CoreBundle] Let the user configure the tmp directory


ngodfraind:
Instead of using sys_get_temp_dir everywhere.

CPASimUSante:
Would that be a terrible idea to have a /claroroot/upload/tmp or claroroot/files/tmp ? they have to be write enabled.

stefk:
@ngodfraind, what's the problem with the system tmp dir? AFAIK even if we don't use it, other libs might do it anyway. Plus it's cleaned by the OS, that's one thing we don't have to take care of.

@CPASimUSante, we can't use the upload dir because it's public. The files dir should also be writable.

ngodfraind:
On most linux OS's (it's what every debian/ubuntu I ever used did) it's only cleaned after a reboot. It's ok a for a dev computer, but for a server it's a bit problematic. This is a quite a common request actually.

CPASimUSante:
@stefk i know they are public folder. But aren't those uploaded files (csv user list, WS...) temp files, wich are deleted (?) automatically after the corresponding script is done? But yeah, for certainty i agree a totally  unreachable dir would be best

stefk:
@ngodfraind, good point.

> But aren't those uploaded files (csv user list, WS...) temp files, wich are deleted (?)

I hope so ;)

charlotte42:
Currently our / tmp directory on the production server is filled quickly. We had to put a script that cleans every 30 minutes ...

PR #1446, opened 2016-10-19
[Distribution] Temporary folder

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Fixed tickets | #1065


Elorfin:
It would be better to remove your merge commit and do a rebase instead if you want to keep your history.

ngodfraind:
It's ok, we can override it.

ngodfraind:
I'm using git hooks and... it doesn't format properly.
@stefk

Elorfin:
I've add the same problem multiple time. In fact, the fix fails each time there is comment block before the `namespace` statement.

LaurentGruber:
LGTM

kitan1982:
LGTM

LaurentGruber:
LGTM

PR #1466, opened 2016-10-21
[Distribution] Load with webpack

ngodfraind:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | no |
| BC breaks? | no |

Work in progress. There is a lot of work and it will require a lot of testing because it's error prone due to the sheer quantity of changed files.
- [x] core 
- [x] activity-tool
- [x] agenda
- [x] announcement
- [x] audio-recorder
- [x] badge
- [x] chat - *Strophe still need to be imported properly*
- [x] clacoform
- [x] collecticiel - *popup.js file is a complete mess and should be tested*
- [x] competency
- [x] cursus - *Should be tested because a lot of code was removed*
- [x] dashboard
- [x] dropzone
- [x] favourite
- [x] flashcard
- [x] formula - *requires #1537*  @ptsavdar  
- [x] forum
- [x] image-player
- [x] ldap - *Should be tested*
- [x] lesson - *Should be cleaned: many useless html.twig files*
- [x] media-resource
- [x] message
- [x] notification
- [x] o365
- [x] oauth
- [x] path - *Tags type ahead; probably broken currently*
- [x] pdf-generator
- [x] pdf-player
- [x] portfolio - *requires testing and has 3rd party libraries... cleaning needed*.
- [x] presence - *it has 3rd party libraries in its public/js directory. Requires tests*.
- [x] reservation
- [x] result
- [x] rss-reader
- [x] scorm
- [x] social-media
- [x] support
- [x] survey
- [x] tag
- [x] team
- [x] text-player
- [x] url
- [x] video-player
- [x] video-recorder
- [x] web-resource
- [x] website
- [x] wiki @dtisch 

TODO:
remove references to the following globals:
They require #1537 to be done.

- [x] tinyMCE
- [x] tinymce
- [x] initTinyMCE
- [x] stfalcon_tinymce_config
- [x] window.tinymce
- [x] restore mention plugin
- [x] remove stfalcon bundle
- [x] restore tinymce locale
- [x] tinymce website
- [x] tinymce lesson
- [ ] tinymce css shouldn't be affecting the whole page //should be fine but I'm not sure
- [x] tinymce lightgray shouldn't trigger 404 error

Chunks:

- [x] use the chunks method everywhere
- [x] remove ng-app directive and include it into the chunk initialisation

Scorm:
- [ ] remove hotAsset from scorm
- [ ] remove ng-app from scorm 

Externals that should be renamed:

- [x] angular
- [x] angular-dragula
- [x] angular-data-table
- [x] at-table
- [x] angular-strap
- [x] angular-bootstrap
- [x] moment
- [x] lodash
- [x] jquery-ui
- [x] confirm-bootstrap
- [x] Datejs
- [x] wavesurfer
- [x] fullcalendar
- [x] pdfjs

Known errors:
- [x] daterangepicker in portfolio 
- [x] datetimepicker in agenda
- [x] loadjs refresh 
- [x] innova video/audio resource doesn't work anymore (was related to loadjs refresh)
- [x] data-table is broken (was related to the installation of angular 1.6.0 - angular 1.5.6 is working)
- [x] js errors in badge managament (admin/badges/edit)
- [ ] top bar in resource manager doesn't work

TODO:
- [x] trim composer.json dependencies
- [x] be careful about the inno angular front end dep
- [ ] remove import from asset('packages'...)
- [ ] check badge jqPlot ?
- [x] typeahead css for competencies

FRONTENDBUNDLE:
- [x] badges
- [ ] blog @dtisch
- [x] chat
- [x] competency
- [x] cursus - **require testing**
- [ ] exo @UJM-dev 

DEPENDENCIES TO CHECK:
- [x]  jqPlot

DISTRIBUTION
- [x] remove js vendors from presence
- [x] remove js vendors from portfolio


Elorfin:
Do you have also modified the SCORM export (for resources that implement it) ?
In the Resource Listener, there is a method `onScormExport` which uses the path to files in the `web` directory.

If it's not updated to use webpack instead, the export will be broken (example with webpack : https://github.com/claroline/Distribution/blob/master/plugin/exo/Listener/ExerciseListener.php#L221).

ngodfraind:
not yet, feel free to do it if you want to

Elorfin:
I will when I will find some time to do it.
I just wanted to write it somewhere to not forget it.

ptsavdar:
@ngodfraind @stefk I tried to test the code in prod environment and there was not much that worked. This was due to the fact that “commons.js” script was loaded after some webpack entries. Though this is not an issue in dev environment since there is no common.js file, in prod however it broke all the code. So I tried to place it higher in the javascripts.html.twig file, only to find out that some dependencies necessary to commons.js were loaded afterwards. So I did a cleanup in the javascripts.html.twig file and most of the code worked. The biggest issue though comes when a webpack script is loaded inside a twig template BEFORE the javascripts.html.twig are loaded (they are loaded in the footer). A solution would be to load all these necessary javascript files inside the head tag in the beginning of the document but then we lose in performance. Trying to find a solution I came up with a workaround for this problem. I created a twig extension WebpackJavascripstExtension.php that allows via a filter (appendScript) to store the script in a table and load it inside javascripts.html.twig after all other scripts have been loaded, using a twig method (loadAppendedScripts()). However if the form is loaded through an AJAX call it loads the script directly and it doesn’t append it to the array (assuming that if it is loaded through AJAX the rest of the page is already there). I tested in both prod and dev environment and it seems to work. I still haven’t changed/tested everything but for the moment I changed all SF2 custom form fields that load webpack scripts and the notification dropdown template. I’ll continue my tests next week. I don’t know if this is the best or even a good solution/workaround but it was the only thing I could think of. I’m open to suggestions if anyone has some other vision of the matter.

stefk:
@ptsavdar, if that problem happens with commons.js dependencies, there's a good chance it will occur with dlls too. I think the main problem here is having dynamically loaded scripts executing before the page content is fully loaded. Shouldn't we consider this as an error? As we're getting heavier on client side features, missing dependencies might not be the only problem. We might need a whole context being properly initialized before doing any work. What do you think? Is delaying ajax requests until page body is fully loaded an option? Will it require too much work?

PS: as for async dependency resolution, we'll probably have a solution with the full es6 `System` object. It should be available in webpack 2 ([ref](https://gist.github.com/sokra/27b24881210b56bbaff7#code-splitting-with-es6)).

ngodfraind:
For the record, I'm currently rewriting most of the tinymce js file because we are using many different instances of it (stfalcon & angular). 

Also it's extendable and the system changed heavily. This will need carefull review but you won't be able to see the diff easily - too many changes with the other files. I don't really care about most of the other files as long as they still work properly, but tinymce is a key point.

ptsavdar:
@stefk The main problem is not for dynamically loaded scripts (through ajax) but for scripts "injected" into pages inside the container or the body BEFORE the footer javascripts. A typical example is the one with SF2 form fields such as TwoLevelSelectType, UserPickerType etc. which embed a webpack entry js file the moment they are inserted in a form.

ptsavdar:
@ngodfraind I agree for tinymce.js file. I haven't changed anything there. Just making sure the platform works properly in prod environment

stefk:
@ptsavdar ok, dynamic or not, that's not the point, but I think the idea is still relevant. We've got code being executed "too soon", out of context. Without any mean to express the dependency directly and force a correct execution order (like in `System.import('foo').then(() => {})`) we could simply delay the execution of that code with an `onload` listener.

ptsavdar:
@stefk In case we want to delay the execution of the code, yes we could add a customized "onload" listener who will catch a custom event. However, there is still the problem for the "webpackJsonP" not defined if commons.js is not already loaded. Because since we're using the CommonChunksPlugin all webpack scripts for production need commons.js to be already loaded in order to work (even with listener) is a webpack automated mechanism/procedure.

stefk:
@ptsavdar right, this one is a real a problem. I guess we can't have a really clean solution until `System.import` is available, so your workaround could do the the trick. However, it seems that webpack's [require.ensure](https://webpack.github.io/docs/code-splitting.html#commonjs-require-ensure) could achieve a similar result, it could be worth trying.

ngodfraind:
@claroline/admins 

This branch is now ready to be tested (prod & dev). Here is the current state.

Known broken stuff:
- [ ] the angular version of the blog. The current one is probably broken with all these changes.
- [ ] the angular version of the exercises. The current one was broken with all theses changes too.
- [ ] maybe a css fix somewhere because it looks wrong to me.
- [ ] the scorm exports should still be changed

Required npm dependencies:
```
    "exports-loader": "^0.6.3",
    "imports-loader": "https://registry.npmjs.org/imports-loader/-/imports-loader-0.6.5.tgz",
    "copy-webpack-plugin": "^4.0.1"
```

The package.json file was changed (see the scripts)

```
{
  "name": "claroline",
  "version": "1.0.0",
  "description": "Claroline dev dependencies and scripts",
  "scripts": {
    "bower": "node_modules/bower/bin/bower update --allow-root",
    "themes": "node vendor/claroline/distribution/main/core/Resources/server/scripts/build-themes",
    "old-themes": "node vendor/claroline/distribution/main/core/Resources/server/scripts/build-old-themes",
    "uglify": "for f in web/dist/*.js; do echo \"Uglifying ${f}...\" && ./node_modules/.bin/uglifyjs $f -o $f; done",
    "webpack": "export NODE_ENV=production && rm -rf web/dist/* && npm run externals && node_modules/webpack/bin/webpack.js  --json --config=vendor/claroline/distribution/webpack.prod.js --progress --colors --display-error-details > webpack-stats.json && npm run uglify",
    "watch": "rm -rf web/dist/* && npm run externals && node_modules/webpack-dev-server/bin/webpack-dev-server.js --hot  --config=vendor/claroline/distribution/webpack.dev.js --inline --content-base web --claro-tgt=watch --display-error-details",
    "karma": "node_modules/.bin/karma start vendor/claroline/distribution/karma.conf.js",
    "externals": "node_modules/webpack/bin/webpack.js --config=vendor/claroline/distribution/webpack.externals.js --progress --colors --display-error-details",
    "mocha": "./node_modules/mocha/bin/mocha vendor/claroline/distribution/main/core/Resources/server/**/*.test.js"
  },
  "dependencies": {
    "claroline-distribution": "file:./vendor/claroline/distribution"
  },
  "devDependencies": {
    "assets-webpack-plugin": "^3.4.0",
    "autoprefixer": "^6.3.1",
    "babel-core": "^6.5.1",
    "babel-loader": "^6.2.2",
    "babel-plugin-transform-runtime": "^6.5.0",
    "babel-preset-es2015": "^6.5.0",
    "babel-preset-react": "^6.11.1",
    "bower": "^1.7.7",
    "colors": "^1.1.2",
    "copy-webpack-plugin": "^4.0.1",
    "css-loader": "^0.25.0",
    "cssnano": "^3.4.0",
    "eslint": "^2.13.1",
    "exports-loader": "^0.6.3",
    "file-loader": "^0.9.0",
    "filesystem-bower-resolver": "^1.2.0",
    "imports-loader": "https://registry.npmjs.org/imports-loader/-/imports-loader-0.6.5.tgz",
    "karma": "^0.13.22",
    "karma-chrome-launcher": "^1.0.1",
    "karma-mocha": "^1.0.1",
    "karma-webpack": "^1.7.0",
    "less": "^2.5.3",
    "mocha": "^2.5.3",
    "postcss-cli": "^2.5.0",
    "raw-loader": "^0.5.1",
    "redux-mock-store": "^1.1.4",
    "shelljs": "^0.5.3",
    "style-loader": "^0.13.1",
    "uglify-js": "^2.6.2",
    "url-loader": "^0.5.7",
    "webpack": "^1.12.13",
    "webpack-dev-server": "^1.14.1",
    "webpack-fail-plugin": "^1.0.4"
  }
}
```

PR #1490, opened 2016-10-26
[exo-v2] API updates

Elorfin:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no

- Adds Pair, Set, Match types (serialization is WIP)
- Migrates old Match data (needs a plugin update)
- Fixes QTI import (this are just quick fixes, the whole system needs to be updated)

### DIFF
--- a/plugin/exo/Tests/Serializer/StepSerializerTest.php
+++ b/plugin/exo/Tests/Serializer/StepSerializerTest.php
@@ -1,7 +1,49 @@
 <?php
-/*
- * Created by PhpStorm.
- * User: Corum
- * Date: 26/09/2016
- * Time: 00:49
- */
+
+namespace UJM\ExoBundle\Tests\Serializer;
+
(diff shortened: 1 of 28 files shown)


plugin/exo/Tests/Serializer/StepSerializerTest.php
```diff
@@ -1,7 +1,49 @@
...
+    {
+        parent::setUp();
+
+        // We trust validator service as it is fully tested
+        $this->validator = $this->client->getContainer()->get('ujm_exo.validator.step');
+        $this->serializer = $this->client->getContainer()->get('ujm_exo.serializer.step');
+    }
+
+    public function testSerializedDataAreSchemaValid()
+    {
```
stefk: You should use `markTestIncomplete` instead of just leaving the test empty. It would help keeping track of what tests remain to be implemented. https://phpunit.de/manual/current/en/incomplete-and-skipped-tests.html

PR #1513, opened 2016-11-02
Move dev deps to project root + remove old scripts

stefk:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no

See https://github.com/claroline/Claroline/pull/996. The new list should be extended with any dependency used exclusively with webpack (and that dependency should then be removed from bower.json). 

**EDIT**: also remove pre-dist scripts and warning


stefk:
LGTM

LaurentGruber:
LGTM

PR #1514, opened 2016-11-03
[ExoBundle] V2 Object Picker

pitrackster:
Add object picker modal

For now it's only a resource picker


plugin/exo/Resources/modules/editor/components/object-picker.jsx
```diff
@@ -0,0 +1,63 @@
...
+      callback: (nodes) => {
+        this.addResource(nodes)
+        // Remove checked nodes for next time
+        nodes = {}
+      }
+    }
+  }
+
+  addResource(nodes){
+    if (typeof nodes === 'object' && nodes.length !== 0) {
```
stefk: Are you sure `length` is meaningful here (type is object, not array)?

plugin/exo/Resources/modules/editor/components/object-picker.jsx
```diff
@@ -0,0 +1,63 @@
...
+        <a role="button" onClick={this.togglePicker.bind(this)}>
+          <i className="fa fa-folder-open"></i>&nbsp;{t('add_resource')}
+        </a>
+      </div>
+    )
+  }
+}
+
+ObjectPicker.propTypes = {
+  onExitedObjectPicker: T.func.isRequired,
```
stefk: This is more `onSelect` than `onExit` IMO.

plugin/exo/Resources/modules/editor/components/object-picker.test.jsx
```diff
@@ -0,0 +1,33 @@
...
+      <ObjectPicker
+        title={123}
+        onExitedObjectPicker="foo"
+      />
+    )
+    ensure.invalidProps('ItemForm', [
+      'title',
+      'onExitedObjectPicker'
+    ])
+  })
```
stefk: A very basic render test would be useful (the resource manager itself could be mocked if needed).

plugin/exo/Resources/modules/editor/components/object-picker.jsx
```diff
@@ -0,0 +1,63 @@
...
+import {t} from './../lib/translate'
+
+/* global Claroline */
+
+class ObjectPicker extends Component {
+  constructor(props){
+    super(props)
+    this.resourcePickerParams = {
+      isPickerMultiSelectAllowed: false,
+      typeWhiteList: [ 'file' ],
```
Elorfin: This needs to be linked to a component prop.

plugin/exo/Resources/modules/editor/components/object-picker.jsx
```diff
@@ -0,0 +1,63 @@
...
+const PICKER_NAME = 'EXO_OBJECT_PICKER'
+import {t} from './../lib/translate'
+
+/* global Claroline */
+
+class ObjectPicker extends Component {
+  constructor(props){
+    super(props)
+    this.resourcePickerParams = {
+      isPickerMultiSelectAllowed: false,
```
Elorfin: This needs to be linked to a component prop.

plugin/exo/Resources/modules/editor/components/object-picker.jsx
Elorfin: The object `nodes` contains : `{nodeId1: node1, nodeId2: node2, ...}`. That's why there is this test.
But the `for` loop can do the job as well.

plugin/exo/Resources/modules/editor/components/object-picker.jsx
```diff
@@ -0,0 +1,63 @@
...
+      // Open existing picker
+      Claroline.ResourceManager.picker(PICKER_NAME, 'open')
+    }
+  }
+
+  render(){
+    return(
+      <div>
+        <a role="button" onClick={this.togglePicker.bind(this)}>
+          <i className="fa fa-folder-open"></i>&nbsp;{t('add_resource')}
```
Elorfin: Prefer the use of `<span>` instead of the deprecated `<i>`.

plugin/exo/Resources/modules/editor/components/object-picker.jsx
stefk: `length` is undefined on plain objects, so `nodes.length !== 0` is always true.

PR #1537, opened 2016-11-07
[CoreBundle] Js entry point: first draft.

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no

@stefk 

This is only a draft but if it's ok I'll improve it in order to make it usable (maybe some libs allowing to parse the file easily) and clean things up.

So there are 2 stubs configs files here:

```
function helloWorld() {
    console.error('hello world !')
}

module.exports = {
  administration: {
    users: {
      buttons: [
        {
          name: 'cursus',
          href: 'path/to/cursus',
          icon: 'ahah/je/suis/cursus',
          callback: helloWorld
        }
      ]
    }
  }
}

```

They look like this and are located in .../Resources/config/config.json


ngodfraind:
@stefk This is working as intended but:

- the config file generation is located in the dll step (I didn't manage to put it elsewhere, webpack just crashed everytime).
- there is no hash at the end of the file.
- I don't think these files are watched by webpack (because it's in the dll step).

any idea ?

ngodfraind:
@stefk 

This requires Claroline/Claroline#1006

ngodfraind:
The error happens because Claroline/Claroline#1006 is required

main/core/Resources/modules/_librairy/Configuration/Configuration.test.js
```diff
@@ -0,0 +1,20 @@
+import Configuration from './Configuration'
+import assert from 'assert'
+
+import fullAction from './test-stubs/actions-full'
```
stefk: Don't put the data associated with tests in separate files unless you really need it. Add the data to the test itself or if it's reused across several tests within the same file, simply use a helper function to return it each time you need it.

main/core/Resources/modules/_librairy/Configuration/Configuration.test.js
```diff
@@ -0,0 +1,20 @@
...
+import partialAction from './test-stubs/actions-partial'
+
+it('Create a configuration data structure', () => {
```
stefk: Tests should be wrapped in a `describe` function. `describe` and `it` are two parts of a complete sentence that will show up in reports and failures.

```js
describe('Configuration', () => {
  it('lists users administration actions', () => {

  })

  if('sets default properties of actions if needed', () => {

  })
})
```
Will read: 
```
Configuration lists users administration actions
Configuration sets default properties of actions if needed
```

main/core/Resources/server/webpack/build/configuration/plugin.js
```diff
@@ -0,0 +1,68 @@
+const paths = require('../../paths')
+const entries = require('../../entries')
+const fs = require('fs')
+const validator = require('./validator')
+
+// no implementation yet
+function ConfigurationPlugin (options) {
+  // Setup the plugin instance with options...
+}
```
stefk: These two comments are superfluous, and the `options` argument as well.

main/core/Resources/server/webpack/build/configuration/plugin.js
```diff
@@ -0,0 +1,68 @@
...
+
+//global because I'm lazy atm
+var first = true
```
stefk: This and the `writeln` stuff should be remplaced with proper string templating and `join('\n')` where needed.

main/core/Resources/server/webpack/build/configuration/validator.js
```diff
@@ -0,0 +1,22 @@
...
+            if (!action.hasOwnProperty(property)) {
+                throw new Error(`The property ${property} is required for each action`)
+            }
+        })
+    })
+}
+
+module.exports = {
+  validate
+}
```
stefk: A lot of 4 vs 2 spaces in this file.

main/core/Resources/server/webpack/build/configuration/validator.test.js
```diff
@@ -0,0 +1,13 @@
+const validator = require('./validator')
+const assert = require('assert')
+const actionFail = require('./test-stubs/actions-fail')
+const actionPass = require('./test-stubs/actions-pass')
+
+it('Test the action validation process', () => {
```
stefk: Same remark than above.

main/core/Resources/server/webpack/plugins.js
```diff
@@ -117,14 +132,16 @@ const dlls = () => {
...
-      context: '.',
-      manifest
-    })
+  return manifests.map(manifest => new webpack.DllReferencePlugin({
+    context: '.',
+  manifest})
   )
 }
 
+const clarolineConfiguration = () => {
```
stefk: Better name needed.

main/core/Resources/modules/_librairy/Configuration/Configuration.js
```diff
@@ -0,0 +1,64 @@
+import config from 'clarolineconfig'
+import union from 'lodash/union'
+import get from 'lodash/get'
+
+class Configuration {
+  constructor (config) {
```
ngodfraind: can you enforce it for travis ?

stefk: Yup: http://eslint.org/docs/rules/space-before-function-paren

main/core/Resources/modules/_librairy/Configuration/Configuration.test.js
ngodfraind: I'm not sure about that one. The test files will be very large because of this eventually. It's ok now because there are only 2 tests but once we'll add more there might be way to many configs and it'll eventually become unreadable.

ngodfraind:
@stefk This is now blocking for #1466

stefk:
LGTM

ngodfraind:
LGMT

main/core/Resources/server/webpack/entries.js
```diff
@@ -123,36 +126,50 @@ function extractPackageInfo(rootDir) {
...
+  return bundles
 }
 
-function isMetaPackage(rootDir) {
+function isMetaPackage (rootDir) {
   return fs.existsSync(rootDir + '/main')
 }
 
-module.exports = collectEntries
+module.exports = module.exports = {
```
Elorfin: Duplicate `module.exports`.

Elorfin:
LGTM

PR #1538, opened 2016-11-08
updates sensio configuration

Elorfin:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no

- Excludes `.travis.yml` from checks
- Removes deleted files from configuration file
- Reduces the constraints on PRs to only validate on `critical` errors. We have too many `major` in the current code base, so the check will always fail. We will increase the severity when the code will be cleaner.

### DIFF
--- a/.sensiolabs.yml
+++ b/.sensiolabs.yml
@@ -1,10 +1,12 @@
+commit_failure_conditions:
+  - "project.severity.critical > 0"
+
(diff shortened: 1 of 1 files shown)


stefk:
LGTM

Elorfin:
LGTM

PR #1545, opened 2016-11-08
[ClarolineCore] es6 shim not loaded with IE11

pitrackster:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no

Allow ES6 features to work with IE11 and other non compatible browsers


LaurentGruber:
LGTM

stefk:
LGTM

PR #1547, opened 2016-11-09
[ExoBundle] Editor v2 open question

pitrackster:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

Add open question to editor
Player is still missing

### DIFF
--- a/plugin/exo/Resources/modules/editor/items/open.js
+++ b/plugin/exo/Resources/modules/editor/items/open.js
@@ -1,32 +1,75 @@
 import {Open as component} from './open.jsx'
 import {ITEM_CREATE} from './../actions'
-import {update} from './../util'
 import {notBlank, number, gteZero} from './../lib/validate'
+import {makeActionCreator} from './../util'
+import cloneDeep from 'lodash/cloneDeep'
+import merge from 'lodash/merge'
+import set from 'lodash/set'
+
+const UPDATE_OPEN = 'UPDATE_OPEN'
+
+export const actions = {
+  updateOpen: makeActionCreator(UPDATE_OPEN, 'id', 'property', 'value')
+}
 
 function reduce(item = {}, action) {
   switch (action.type) {
     case ITEM_CREATE: {
-      return update(open, {
-        maxScore: {$set: 0},
-        maxLength: {$set: 0}
+      return Object.assign({}, item, {
+        maxScore: 0,
+        maxLength: 0
       })
+    }
+
+    case UPDATE_OPEN: {
+      const newItem = cloneDeep(item)
+      newItem._touched = merge(
+        newItem._touched || {},
+        set({}, action.property, true)
+      )
+      const newValue = parseFloat(action.value)
+      if(action.property === 'maxScore'){
+        newItem.maxScore = newValue
+      } else if(action.property === 'maxLength'){
+        newItem.maxLength = newValue
+      }
+      return newItem
     }
   }
   return item
 }
 
-// function initialFormValues(item) {
-//   return update(item, {
-//     maxScore: {$set: item.score.max}
-//   })
-// }
 
 function validate(values) {
-  return {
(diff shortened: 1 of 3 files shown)


plugin/exo/Resources/modules/editor/items/open.js
```diff
@@ -1,32 +1,75 @@
...
 import {notBlank, number, gteZero} from './../lib/validate'
+import {makeActionCreator} from './../util'
+import cloneDeep from 'lodash/cloneDeep'
+import merge from 'lodash/merge'
+import set from 'lodash/set'
+
+const UPDATE_OPEN = 'UPDATE_OPEN'
+
+export const actions = {
+  updateOpen: makeActionCreator(UPDATE_OPEN, 'id', 'property', 'value')
```
stefk: `id` doesn't seem to be used anywhere.

plugin/exo/Resources/modules/editor/items/open.js
```diff
@@ -1,32 +1,75 @@
...
+      newItem._touched = merge(
+        newItem._touched || {},
+        set({}, action.property, true)
+      )
+      const newValue = parseFloat(action.value)
+      if(action.property === 'maxScore'){
+        newItem.maxScore = newValue
+      } else if(action.property === 'maxLength'){
+        newItem.maxLength = newValue
+      }
```
stefk: The 6 lines above can be reduced to : 
```js
newItem[action.property] = parseFloat(action.value)
```

plugin/exo/Resources/modules/editor/items/open.js
```diff
@@ -1,32 +1,75 @@
 import {Open as component} from './open.jsx'
 import {ITEM_CREATE} from './../actions'
-import {update} from './../util'
 import {notBlank, number, gteZero} from './../lib/validate'
+import {makeActionCreator} from './../util'
+import cloneDeep from 'lodash/cloneDeep'
+import merge from 'lodash/merge'
+import set from 'lodash/set'
+
+const UPDATE_OPEN = 'UPDATE_OPEN'
```
stefk: These actions are already isolated by question types so I don't think you need to use `OPEN` in it. `UPDATE` or `UPDATE_PROPERTY` should be enough.

plugin/exo/Resources/modules/editor/items/open.js
```diff
@@ -1,32 +1,75 @@
...
+    errors = {
+      maxLength : number(values.maxLength)
+    }
+  } else if(undefined !== gteZero(values.maxLength)){
+    errors = {
+      maxLength : gteZero(values.maxLength)
+    }
   }
+  return errors
 }
```
stefk: Look at [this](https://github.com/claroline/Distribution/blob/exo-v2/plugin/exo/Resources/modules/editor/validators.js#L10) for an example of how to reduce that boilerplate.

plugin/exo/Resources/modules/editor/items/open.jsx
```diff
@@ -1,17 +1,51 @@
-import React from 'react'
-import {Field} from 'redux-form'
+import React, {PropTypes as T}from 'react'
```
stefk: Missing space after `}`

plugin/exo/Resources/modules/editor/items/open.test.js
```diff
@@ -0,0 +1,176 @@
...
+        item={{
+          id: '1',
+          content: 'Question?',
+          maxLength: 255,
+          maxScore: 10
+        }}
+        onChange={() => {}}
+      />
+    )
+    ensure.propTypesOk()
```
stefk: Testing prop types are ok isn't enough IMO. At least we could test inputs have been properly rendered. A complete test would also check that we get the data back when an input is filled.

[that test is also missing for choice.js, I simply didn't have the time to do it]

PR #1411, opened 2016-10-14
[AnnouncementBundle] Adds workspace code and name in sent mail

kitan1982:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Fixed tickets | #1276 

- Workspace code has been added in the mail object
- Workspace code and name have been added at the end of the mail content

PR #1553, opened 2016-11-14
[AnnouncementBundle] Add workspace name and code to announcement message

stefk:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no

https://github.com/claroline/Distribution/pull/1411 has added workspace code and name to the content of announcements, but only when users are picked as explicit recipients. For usual announcements which transit through the message system, that infomation is missing.

Notes: 
- the "fix" here is just a copy/paste of the layout chosen by #1411 which is really ugly
- maybe that kind of information need should be handled at a higher level


LaurentGruber:
LGTM

kitan1982:
LGTM

PR #1563, opened 2016-11-15
[CoreBundle] Fixing resource soft delete

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no


LaurentGruber:
LGTM

main/core/Manager/ResourceManager.php
```diff
@@ -1014,26 +1014,26 @@ public function delete(ResourceNode $resourceNode, $force = false)
...
+                        'claroline_resources_delete',
+                        'GenericDatas',
+                        [[$node]]
+                    );
+
+                    $this->dispatcher->dispatch(
+                        'log',
+                        'Log\LogResourceDelete',
+                        [$node]
+                    );
```
Elorfin: Why not dispatching the log for soft deletes ? 

IMHO, how we delete the data has no concern with the result : for the end user, the resource is deleted, so it needs to be notified as is.

main/core/Manager/ResourceManager.php
```diff
@@ -1014,26 +1014,26 @@ public function delete(ResourceNode $resourceNode, $force = false)
...
                 if ($softDelete || $eventSoftDelete) {
                     $node->setActive(false);
                     $this->om->persist($node);
                 } else {
+                    //what is it ?
+                    $this->dispatcher->dispatch(
+                        'claroline_resources_delete',
+                        'GenericDatas',
+                        [[$node]]
+                    );
```
ngodfraind: I did a quick search on this listener. It's used by the TagListener.
It removes the tags linked to the resources.

I didn't know it was here but I kind of understand why it exists as it allows to retrieve every resources at the end of the deletion process (if it's a directory) and loop through them.
The result would have been the same if the it listened to the delete event but in the current build, it can't easily do this because the name changes for each resources.

main/core/Manager/ResourceManager.php
ngodfraind: Agree

PR #1580, opened 2016-11-21
[ExoBundle] Words question form minor changes

pitrackster:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no

- Add negative-score positive-score css classes depending on the item score
- Replace keyword textarea field with simple text input
- Translations

### DIFF
--- a/plugin/exo/Resources/modules/editor/items/words.jsx
+++ b/plugin/exo/Resources/modules/editor/items/words.jsx
@@ -16,14 +16,16 @@ class WordItem extends Component {
 
   render() {
     return (
-      <div className="word-item">
+      <div className={classes('word-item', {'positive-score': this.props.score > 0 } , {'negative-score': this.props.score <= 0 })}>
         <div className="text-fields">
-          <Textarea
+          <input
+            type="text"
             id={`solution-${this.props.index}-text`}
             title={tex('response')}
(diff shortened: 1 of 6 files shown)


plugin/exo/Resources/modules/editor/items/words.jsx
```diff
@@ -16,14 +16,16 @@ class WordItem extends Component {
 
   render() {
     return (
-      <div className="word-item">
+      <div className={classes('word-item', {'positive-score': this.props.score > 0 } , {'negative-score': this.props.score <= 0 })}>
```
Elorfin: This may be interesting to have a component for this. So we can have a common interface to use for all kind of question items.

plugin/exo/Resources/modules/editor/style.css
```diff
@@ -174,7 +174,7 @@
   top: -5px;
   right: -5px;
   z-index: 1;
-  background-color: white;
+  background-color: transparent;
```
Elorfin: that was to avoid seeing the corner of the input in the icon (inside the + sign). If you want to fix the display, you can just apply a `border-radius: 50%;` to avoid the white square around.

PR #1581, opened 2016-11-21
[ExoBundle] Question picker

pitrackster:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no

Search and import questions while editing / creating an exercise


plugin/exo/Resources/modules/editor/actions.js
```diff
@@ -3,6 +3,7 @@ import select from './selectors'
 import {makeActionCreator, makeId} from './util'
 
 export const ITEM_CREATE = 'ITEM_CREATE'
+export const ITEM_CREATE_FROM_EXISTING = 'ITEM_CREATE_FROM_EXISTING'
```
stefk: `ITEM_IMPORT` would probably be better.

plugin/exo/Resources/modules/editor/actions.js
```diff
@@ -56,6 +64,16 @@ actions.createItem = (stepId, type) => {
   }
 }
 
+actions.createItemFromExisting = (stepId, questions) => {
+  invariant(stepId, 'stepId is mandatory')
+  invariant(questions, 'questions is mandatory')
+  return {
+    type: ITEM_CREATE_FROM_EXISTING,
+    stepId,
+    items: questions
```
stefk: Why not naming the argument `items` and using `makeActionCreator`?

PR #1588, opened 2016-11-22
User password update via csv

kitan1982:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no

Updates password of existing users during csv import only if password is not empty


ptsavdar:
LGTM

LaurentGruber:
LGTM

PR #1589, opened 2016-11-22
[CoreBundle] Create default organization if it doesn't already exist

ptsavdar:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Fixed tickets | 

Creates default organization in case there is none. (avoid having bug when creating user account using default organization)


LaurentGruber:
LGTM

purplefish32:
LGTM

PR #1590, opened 2016-11-22
Fix category filter and add translations to notifications

CPASimUSante:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Fixed tickets | https://github.com/claroline/Distribution/issues/1550

It seems the values of the data to be translated near the icon are set in each bundle for each notification event with the getIconKey() string for the Event : for now different event use the same string key for a bundle (CommentCreate, CommentPublish, CommentUpdate... for the blog as en example). For the Core Bundle, there is no key set. Maybe in the future we could imagine having more specific keys (like blog-comment-add, forum-question-add)


LaurentGruber:
LGTM

Elorfin:
LGTM

PR #1652, opened 2016-12-11
fix display and button behavioral button problems on the team page

CPASimUSante:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no ?
| Fixed tickets | https://github.com/claroline/Distribution/issues/1478

This fixes 2 things (+1) : 
- the "register" button is unclickable or not present if the max register user is reached.
- the display of the team page during the impersonnation was wrong

Fixing the team bundle display, it occured that there was also a behavior bug for the administrator when impersonating. It is fixed.

Maybe a build could be set before validating.

@ngodfraind : is the fix in the voter OK ?
@ any reviewer : could my fix in plugin/team/Resources/views/Team/userMenu.html.twig be simpler ?

the condition (i fixed), is now : 
```
{% if team.getMaxUsers() is empty
    or (team.getMaxUsers() is not empty
        and (
            (nbUsers[team.getId()] is defined and (team.getMaxUsers() < nbUsers[team.getId()]))
            or (nbUsers[team.getId()] is not defined and team.getMaxUsers() > 0)
        )
    ) %}
 {% set registerable = true %}
{% endif %}
```
With words : user should be allowed to register
- if there is not a team max user set (team.getMaxUsers())
otherwise, (if the max user is set), 
  * if the number of user in the team is set and the team max user is greater than the number of user in the team
  * or if the number of user in the team is NOT set and the team max user is greater than 0


plugin/team/Resources/views/Team/userMenu.html.twig
```diff
@@ -217,8 +217,21 @@
...
+                                    {% set registerable = true %}
+                                {% else %}
+                                    {% if (nbUsers[team.getId()] is defined) %}
+                                        {% if (team.getMaxUsers() < nbUsers[team.getId()]) %}
+                                            {% set registerable = true %}
+                                        {% endif %}
+                                    {% else %}
+                                        {% if team.getMaxUsers() > 0 %}
+                                            {% set registerable = true %}
+                                        {% endif %}
```
LaurentGruber: dans tout les cas tu auras registerable à true en faisant comme ça non ?

CPASimUSante: Pour moi, registerable ne sera pas true (pas défini)

SI  (team.getMaxUsers() is not empty ET  (nbUsers[team.getId()] is NOT defined) ET (team.getMaxUsers() >= nbUsers[team.getId()]) )

OU
SI (team.getMaxUsers() is not empty ET  (nbUsers[team.getId()] is NOT defined) ET team.getMaxUsers() = 0)

Mais comme j'ai indiqué, il y a peut-être plus simple. (et cette logique pourrait p e se trouver dans le controller...)

ptsavdar: IMO you could replace it with:
```
{% if team.getMaxUsers() is empty OR (nbUsers[team.getId()] is defined AND team.getMaxUsers() < nbUsers[team.getId()]) OR (nbUsers[team.getId()] is not defined AND team.getMaxUsers() > 0 %}
    {% set registerable = true %}
{% endif %}
```

If I got right what you want to accomplish.

CPASimUSante: i think this is more like : 
```
{% if team.getMaxUsers() is empty
    or (team.getMaxUsers() is not empty
        and (
            (nbUsers[team.getId()] is defined and (team.getMaxUsers() < nbUsers[team.getId()]))
            or (nbUsers[team.getId()] is not defined and team.getMaxUsers() > 0)
        )
    ) %}
 {% set registerable = true %}
{% endif %}
```
With words : user should be allowed to register:
- if there is not a team max user set (team.getMaxUsers())
otherwise, (if the max user is set), 
  * if the number of user in the team is set and the team max user is greater than the number of user in the team
  * if the number of user in the team is NOT set and the team max user is greater than 0

PR #1671, opened 2016-12-13
removes lessphp

Elorfin:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no

Removes the last uses of LESS PHP and removes the dependency.


Elorfin:
See #1673.

PR #1672, opened 2016-12-14
Use travis full VM (increase available memory)

stefk:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no

Another attempt at solving memory issues on travis. If we use old VMs instead of docker containers, we should have a few more Go available. https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+sudo: required
+
 language: php
 
 php:


stefk:
Closed in favour of #1673

PR #1691, opened 2016-12-19
[ExoV2] Save quiz (open questions)

pitrackster:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no

Can save a quiz containing open questions.


plugin/exo/Resources/modules/quiz/store.js
```diff
@@ -54,7 +57,32 @@ const reducerSwitcher = () => next => action => {
...
+      credentials: 'include',
+      body: JSON.stringify(denormalized)
+    }
+    fetch(url, params)
+     .then(response => {
+       if(!response.ok){
+         // do something with errors...
+         let result = next(action)
+         return result
+       }
```
stefk: Shouldn't you call `next(action)` when response is ok too?

BTW the intermediate variable `result` isn't necessary, you can simply return `next(action)`

PR #1697, opened 2016-12-19
[exo-v2] first player version

Elorfin:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | yes

- Fixes bugs player api.
- Adds missing `maxLength` prop on Open question in API.
- First player version


plugin/exo/Resources/modules/items/components/player.jsx
```diff
@@ -0,0 +1,35 @@
...
+const Hints = () =>
+  <div>
+    This is the hint section
+  </div>
+
+Hints.propTypes = {
+  hints: T.array.isRequired
+}
+
+export class Player extends Component {
```
stefk: Using this kind of definition suggests the component has some state, which is not the case. Why not using the recommended functional form here?

PR #1699, opened 2016-12-20
[exo-v2] enhance attempts

Elorfin:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no

- Fixes player navigation
- Uses local items for playing attempts (old method was to load it from server with the paper)


plugin/exo/Resources/modules/quiz/player/selectors.js
```diff
@@ -1,9 +1,63 @@
...
+/**
+ * Retrieves the picked items for a step.
+ *
+ * @param {object} state
+ *
+ * @returns {array}
+ */
 select.currentStepItems = (state) => {
-  const stepStructure = state.paper.structure.find((step) => step.id === state.currentStep.id)
+  const stepStructure = state.paper.structure.find((step) => step.id === state.currentStep)
```
stefk: You should probably use reselect here. There's no need to perform a `find` unless the current step has changed.

stefk: Actually most selectors in this file could be composed of smaller/memoized selectors. That would be both more readable and more efficient.

PR #1701, opened 2016-12-20
[ExoV2] Update thumbnail step title when needed

pitrackster:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no

When the step title is updated, the thumbnail is also updated


plugin/exo/Resources/modules/quiz/editor/selectors.js
```diff
@@ -21,7 +22,7 @@ const quizThumbnail = createSelector(
   (quiz, current) => {
     return {
       id: quiz.id,
-      title: 'Exercice',
+      title: t('Éxercice'),
```
stefk: Shouldn't it be "exercise" if translated? Same below.

plugin/exo/Resources/modules/quiz/editor/selectors.test.js
```diff
@@ -8,7 +8,7 @@ describe('Thumbnails selector', () => {
     assertEqual(select.thumbnails(fixtureState1()), [
       {
         id: '1',
-        title: 'Exercice',
+        title: 'Éxercice',
```
stefk: Can't be **é**xercice

om64:
Exercise in English, indeed !

PR #1705, opened 2016-12-20
[ExoV2] Decorate quiz with a default step

pitrackster:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no

Add a default step to quiz if none


plugin/exo/Resources/modules/quiz/decorators.js
```diff
@@ -3,12 +3,22 @@ import cloneDeep from 'lodash/cloneDeep'
...
 import {TYPE_QUIZ} from './enums'
+import {makeId} from './../utils/utils'
 
 // augment normalized quiz data with editor state attributes and default values
 // (can be passed an array of sub-decorators for each item mime type)
 export function decorate(state, itemDecorators = {}) {
   const newState = cloneDeep(state)
 
+  // create en empty step if none
+  if(Object.keys(newState.quiz.steps).length === 0){
```
stefk: `newState.quiz.steps` is already an array, so `Object.keys` is not needed.

Also please use appropriate spaces in that kind of statement: `if (...) {`

PR #1718, opened 2016-12-21
[ExoV2] Change item creation button and item import behavior

pitrackster:
- Change the (Add / Import) question button 
- For now it does not record the last user choice in localstorage
- When openeing import item modal questions are not loaded, the action begins only when the user enter some search critria


plugin/exo/Resources/modules/quiz/editor/components/modals.jsx
```diff
@@ -179,7 +175,7 @@ class ImportItemsModal extends Component {
           { this.state.questions.length === 0 &&
             <div className="text-center">
               <hr/>
-              <h4>{t('no_search_results')}</h4>
+              <h4>{tex('search_question_instructions')}</h4>
```
Elorfin: I'm not sure if we really need to display it. 

Also if your search return no result this is this message which will be displayed. The old trad was better for this.

plugin/exo/Resources/modules/quiz/editor/components/step-editor.jsx
```diff
@@ -158,46 +158,78 @@ ItemPanel.propTypes = {
...
+    // this is required before componentDidMount. If not state is not defined...
+    this.state = {
+      currentLabel: tex('add_question_from_new'),
+      currentAction: MODAL_ADD_ITEM
+    }
+  }
+
+  componentDidMount(){
+    // @TODO get default dropdown action in local storage if any
+  }
```
Elorfin: This sould be removed.

PR #1720, opened 2016-12-21
[CoreBundle] Fixing workspace import 

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no

The current import mechanism didn't use the default value set in the configuration because we never returned the data after the processor fired the processConfiguration method.

It fixes my use case where I needed the default values for the home widgets. I didn't change the rest.

I also slightly simplified the import method in the TransferManager (removes some useless/duplicated code).


ngodfraind:
I messed up with my prs, it's missing some changes. I'm adding them back.

ngodfraind:
done

PR #1751, opened 2017-01-02
[Distribution] Composer

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | maybe

- Update symfony 2.7 branch and required dependencies
- Removes nelmio api doc (barely used and doesn't work with the latest 2.7 (see conflict here: https://github.com/nelmio/NelmioApiDocBundle/blob/master/composer.json)
- Twigjs causes issue with the latest twig branch. A pr already exists but is not merged. This bundle should be removed sooner or later anyway so it's pointing to the pr fork.
https://github.com/schmittjoh/twig.js/pull/109
https://github.com/schmittjoh/twig.js/issues/110

This PR requires claroline/claroline#1038 to work


ngodfraind:
@claroline/admins see the previous message

PR #1757, opened 2017-01-03
[Tinymce] Urls handing

ngodfraind:
absolute urls allowed

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no

https://github.com/tinymce/tinymce-docs/blob/master/_includes/configuration/remove-script-host.md

This is better for sending urls with tinymce by email.

This is a old fix so it should require some testing (I don't remember what it did exactly) but I think it's ok.


main/core/Resources/modules/tinymce/tinymce.js
```diff
@@ -60,11 +60,11 @@ tinymce.claroline.editorChange = function (editor) {
  *  @param args TinyMCE paste plugin arguments.
  *
  */
-tinymce.claroline.paste = function (plugin, args) {
+tinymce.claroline.paste = function(plugin, args) {
   if ($('#platform-configuration').attr('data-enable-opengraph') === '1') {
-    var link = $('<div>' + args.content + '</div>').text().trim() // inside div because a bug of jquery
+    var link = $('<div>' + args.content + '</div>').text().trim // inside div because a bug of jquery
```
stefk: `trim` without function call? Seems weird.

ngodfraind: This is what the linter did (the original change was only the parameter).

LaurentGruber: mais ça risque de bugger sans les ()

PR #1758, opened 2017-01-03
Use param instead of reading store

stefk:

### DIFF
--- a/plugin/exo/Resources/modules/quiz/papers/components/papers.jsx
+++ b/plugin/exo/Resources/modules/quiz/papers/components/papers.jsx
@@ -1,6 +1,6 @@
 import React from 'react'
 
-export const Results = () =>
+export const Papers = () =>
   <div>
-    RESULTS
+    Papers
   </div>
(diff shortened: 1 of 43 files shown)


plugin/exo/Resources/modules/quiz/papers/components/papers.jsx
```diff
@@ -51,20 +52,27 @@ let Papers = props =>
     </thead>
     <tbody>
       {props.papers.map((paper, idx) =>
-        <PaperRow key={idx} admin={props.admin} {...paper}/>
+        <PaperRow
+          key={idx}
+          quizId={props.quizId}
```
Elorfin: I would have passed : `openPaper={(paperId) => props.displayPaper(props.quizId, paperId)}` to simplify the child component and keep it unaware of the quiz.

stefk: It's used to construct an URI, so a callback is not an option.

PR #1759, opened 2017-01-03
[WebSite] changes assetic destination folder

Elorfin:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no

Styles managed by assetic for website are dumped in the same folder used by `asset:installs`.
This produces strange results on my dev install (dumped files also appears in the vendor source directory, website/Resources/public/css, and so in git changes). 

I think it's due to the symlinks (possibly a problem with how windows manages it).

I've just moved the dest folder to another one in the `web` dir.

ping @ptsavdar.


ptsavdar:
@Elorfin My apologies for my late reply. It's good for me. I'm squashing and merging if that's ok for you.

Elorfin:
@ptsavdar : no problem. It was not urgent.

PR #1767, opened 2017-01-04
[ExoV2] Player routing + choice style and fixes

stefk:


Elorfin:
Why have you moved https://github.com/claroline/Distribution/pull/1767/files#diff-5573ea8503d9db6aa4909a82c5d40714L1 ?

I was planning to use this component for the preview in the bank (and in the quiz editor). So IMO it should go in the items module.

We should also expose a root item component like this for the editor and corrections for the same reasons. It would be easier to use.

stefk:
Playing items in the question bank or in the editor isn't a v1 feature. As you said, there're other things we could move in that directory but I don't see the point of doing it now and only partially if that's not needed.

PR #1778, opened 2017-01-06
[ExoV2] Bin feedback icons order

pitrackster:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no

Swap bin and feedback icons on item forms where needed

### DIFF
--- a/plugin/exo/Resources/modules/items/choice/editor.jsx
+++ b/plugin/exo/Resources/modules/items/choice/editor.jsx
@@ -69,20 +69,20 @@ class ChoiceItem extends Component {
                 )}
               />
             }
-            <span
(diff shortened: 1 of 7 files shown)


plugin/exo/Resources/modules/items/choice/editor.jsx
```diff
@@ -71,19 +71,19 @@ class ChoiceItem extends Component {
             }
             <span
               role="button"
```
Elorfin: Just to remind we will need to replace it by `a` or `button` tag for accessibility. As is it's not focusable with a keyboard.

PR #1784, opened 2017-01-09
[ExoV2] Save guard + modals

stefk:


plugin/exo/Resources/less/components/page.less
```diff
@@ -13,6 +13,11 @@
         border-left: none;
         border-right: none;
         margin-bottom: 0;
+
+        button.btn-link {
+          text-decoration: none;
+          outline: none;
```
Elorfin: You should not remove the outline here. All other btn/a in platform have an outline.
This rule must be applied to all the links (after checking it doesn't break accessibility) or not at all.

Issue #1788, opened 2017-01-10
[CoreBundle] Administration Code 'uniqueness' side effects


ptsavdar:
| Q             | A
| ------------- | ---
| Bug report?   | yes
| Version?      | 8.0.2

The use of the administration code in user as a 'unique' field (and thereby as an identification field) has some serious side effects in systems that already used the same code for multiple users. First of all, the fact that there is no 'uniqueness' restriction in administration code (as there is for email and login for example) leaves the possibility to attribute the same code to multiple users. That is the case for many instances that use this code to group together some of their users (it's easier for them to search these users afterwards using this code and as there was no restriction prohibiting this action). The problem is that now, the code is used as an identification field for users, so issue number one: if the username of user X is the same as the administration code of user Y then trying to usurp user X or Y is impossible as there is a 'NoUniqueException' returned from method 'findUserByUsernameOrMailOrCode', used to retrieve the user. That is a minor issue. The big problem comes with account creation via a CSV file. If there are multiple accounts in the CSV file with the same Administration Code ONLY THE LAST ONE IS CREATED, while the message displayed says that every account was created. You can test this with this 
[file](https://github.com/claroline/Distribution/files/695636/FR.PDL.Directeurs.FPC2.txt) (rename it to .csv first). Even worse, if an account with the same code already exists in plateforme it GETS UPDATED with the new user's data even though 'Creation only' option was chosen in the submit form. The result is that many accounts have 'what seems to be' deleted (as the users can no longer log in nor can be found by the search engine) while others (the new accounts) have the content (workspaces, resources etc.) of these 'deleted' accounts. So instead of creating a new account for the CSV user data, an existing account with the same code gets updated to contain these new user data. Which is a big security issue but also a big functional issue.

What would you propose for those instances that have users with the same administration codes? (knowing that changing them manually is not an option as they may have some thousands of users in their DB).

@LaurentGruber @ngodfraind 

Thank you in advance for your replies.

CPASimUSante:
OMG : i do use the same code for many user (even though i should use groups to do this). This issue is then very serious : "changing the rules" (registration mandatory fields, uniqueness or not) when the platforms in prod is indeed a dangerous idea.

ngodfraind:
I'm looking at this.

PR #1790, opened 2017-01-11
[CoreBundle] 8.x fix administration code issue

ptsavdar:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Fixed tickets | #1788 #1638 

- Added option 'is_user_admin_code_unique' in platform parameters to define if administrative code is unique (and thus use it as an identifier) or not. 
- Added unique constraint and validator for administrative code (when it's unique). 
- Fixed user override during CSV import.


main/core/Controller/Administration/UsersController.php
```diff
@@ -353,44 +339,31 @@ public function importAction()
...
+                $additionalRoles,
+                $enableEmailNotification,
+                $options
+            );
 
-                foreach ($updatedNames as $name) {
+            foreach ($logs as $key => $names) {
+                $msgClass = 'success';
+                if ($key === 'skipped') {
+                    $msgClass = 'error';
```
CPASimUSante: isn't it 'danger' ? http://www.w3schools.com/bootstrap/bootstrap_alerts.asp

ptsavdar: it's 'danger' for bootstrap but 'error' for SF2 messages

main/core/Controller/Administration/UsersController.php
```diff
@@ -304,31 +311,10 @@ public function importAction()
...
-                        if (is_null($existingUser)) {
-                            $users[] = $datas;
-                        } else {
-                            $toUpdate[] = $datas;
-                        }
-                    } else {
-                        $users[] = str_getcsv($line, ';');
-                    }
-                }
+                $users[] = str_getcsv($line, ';');
```
CPASimUSante: the update has been skipped in order to avoid checking for code ? So if i understand, now, every user will be updated, not only those already existing ?

ptsavdar: No, it's just that the procedure was already in place in UserManager. So I removed it from UsersController. If "Creation only" is chosen no user gets updated. If "Update" is chosen then already existing users get updated. If admin code is unique and in the CSV file many user have been marked with the same admin code then an error is shown displaying the lines having the same admin code. Basically, the update was skipped because it was already developed in UserManager. To avoid duplication.

ptsavdar:
@CPASimUSante is everything alright for you? Can I merge?

CPASimUSante:
@ptsavdar. Your security checks seems to prevent any displeasant behaviour, so yes for me it's ok.  But i clicked on approved already i think. My question above was not blocking.

ptsavdar:
@CPASimUSante yeap I saw your approval, however I wanted to double-check just to be sure. Thank you for your review and approval.

PR #1795, opened 2017-01-12
[exo-v2] Match question editor

pitrackster:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no

- fixes several problems in editor
- add deserializers
- add Label and Proposal uuid fields
- add match.less file
- also change minor css problems for tooltiped button component


plugin/exo/Resources/modules/items/match/editor.jsx
```diff
@@ -184,11 +204,28 @@ MatchItem.propTypes = {
   onChange: T.func.isRequired
 }
 
+
+const handleRTEClick = function (event) {
```
stefk: This function (and `handleWindowResize`) should be methods of the `Match` class.

stefk: (perhaps a better name too -- no idea what "RTE" means here)

plugin/exo/Resources/modules/items/match/editor.jsx
```diff
@@ -303,6 +381,9 @@ class Match extends Component {
       }
     }.bind(this), 100)
 
+    window.setTimeout(function (){
+      this.jsPlumbInstance.repaintEverything()
+    }.bind(this), 3000)
```
stefk: I think if you use the fat arrow notation (`() => {}`) you don't have to bind `this` (it's done automatically as per the spec).

plugin/exo/Serializer/Question/Type/MatchQuestionSerializer.php
```diff
@@ -51,14 +51,14 @@ public function serialize($matchQuestion, array $options = [])
 
         $firstSet = array_map(function (Proposal $proposal) use ($options) {
             $itemData = $this->contentSerializer->serialize($proposal, $options);
-            $itemData->id = (string) $proposal->getId();
+            $itemData->id = (string) $proposal->getUuid();
```
stefk: The string cast isn't necessary anymore.

plugin/exo/Serializer/Question/Type/MatchQuestionSerializer.php
```diff
@@ -51,14 +51,14 @@ public function serialize($matchQuestion, array $options = [])
...
 
             return $itemData;
         }, $matchQuestion->getProposals()->toArray());
 
         $secondSet = array_map(function (Label $label) use ($options) {
             $itemData = $this->contentSerializer->serialize($label, $options);
-            $itemData->id = (string) $label->getId();
+            $itemData->id = (string) $label->getUuid();
```
stefk: Same.

plugin/exo/Serializer/Question/Type/MatchQuestionSerializer.php
```diff
@@ -78,6 +78,29 @@ public function serialize($matchQuestion, array $options = [])
...
 
+    private function serializeSolutions(MatchQuestion $matchQuestion)
+    {
+        $solutions = [];
+
+        foreach ($matchQuestion->getProposals() as $proposal) {
+            /** @var Label $label */
+            foreach ($proposal->getExpectedLabels() as $label) {
+                $solutionData = new \stdClass();
+                $solutionData->firstId = (string) $proposal->getUuid();
```
stefk: same

plugin/exo/Serializer/Question/Type/MatchQuestionSerializer.php
```diff
@@ -97,35 +120,153 @@ public function deserialize($data, $matchQuestion = null, array $options = [])
             $matchQuestion->setPenalty($data->penalty);
         }
 
-        if (isset($data->shuffle)) {
-            $matchQuestion->setShuffle(true);
+        if (isset($data->random)) {
+            $shuffle = $data->random === 1 ? true : false;
```
stefk: You don't need a ternary operator, `===` is already evaluated to true/false. But why would `data->random` be an integer?

plugin/exo/Serializer/Question/Type/MatchQuestionSerializer.php
```diff
@@ -97,35 +120,153 @@ public function deserialize($data, $matchQuestion = null, array $options = [])
...
-                $solutionData = new \stdClass();
-                $solutionData->firstId = (string) $proposal->getId();
-                $solutionData->secondId = (string) $label->getId();
-                $solutionData->score = $label->getScore();
+        foreach ($secondSets as $index => $secondSetData) {
+            $label = null;
+            // Searches for an existing Label entity.
+            foreach ($secondSetEntities as $entityIndex => $entityLabel) {
+                /** @var Label $entityLabel */
+                if ((string) $entityLabel->getUuid() === $secondSetData->id) {
```
stefk: Cast not needed

PR #1805, opened 2017-01-13
[Exo-v2] Refactoring paper tabs

ngodfraind:
I think there are still some minor display issues (it's hard to test because on my install, some bugs don't allow me to show the Metadata easily).

Anyway, the paper tabs are now a separate React component.


Pauooo:
still few 
- you still have problems with the top of the list in the tabs, even if the description of the question is out of the tabs
captura13
- Negative points are not "working", the score displayed is 0. and in the "your answer" tab, the 0 is not followed by "point"
- pb with the quiz description and the v and the a of "valider" are very close
captura17_01

- new pb with paper list : 
captura17_02

- text and html in popover : 
captura17_04
captura17_03

ngodfraind:
The html will be easily fixed but I'm not so sure about the popover size.

ngodfraind:
As of now

- Can't reproduce anymore the issue where the tab goes out of the panel. I don't know where it come from.

Others issues are fixed afaik.

ngodfraind:
@Pauooo Done

Pauooo:
@ngodfraind still the problem with paper list (but maybe it doesn't concern this pr)  and a new thing: in the feedback, the comments (icon + popover) is available juste for the answer given by the user,not for all of the solutions

PR #1806, opened 2017-01-13
same width for  editor / player / papers / overview

pitrackster:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no

- Harmonize width in player / editor / overview / paper list / paper view
- add variables.less for the bundle

### DIFF
--- a/plugin/exo/Resources/modules/quiz/overview/overview.jsx
+++ b/plugin/exo/Resources/modules/quiz/overview/overview.jsx
@@ -108,7 +108,7 @@ const Layout = props =>
   <div className="quiz-overview">
     {props.empty &&
       <div className="row">
-        <div className="col-md-8 col-md-offset-2">
+        <div className="col-md-12">
           <div className="alert alert-info text-center">
             <span className="fa fa-fw fa-warning"></span>
             <span>
@@ -122,7 +122,7 @@ const Layout = props =>
       </div>
     }
     <div className="row">
-      <div className="col-md-8 col-md-offset-2">
+      <div className="col-md-12">
         {props.description &&
           <div className="exercise-description panel panel-default">
             <div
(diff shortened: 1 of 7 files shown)


plugin/exo/Resources/modules/quiz/overview/overview.jsx
```diff
@@ -108,7 +108,7 @@ const Layout = props =>
   <div className="quiz-overview">
     {props.empty &&
       <div className="row">
-        <div className="col-md-8 col-md-offset-2">
+        <div className="col-md-12">
```
Elorfin: Completely remove the grid here. This is not needed as it's always full width. Tyou can avoid 2 `div` containers.

plugin/exo/Resources/modules/quiz/overview/overview.jsx
```diff
@@ -122,7 +122,7 @@ const Layout = props =>
       </div>
     }
     <div className="row">
-      <div className="col-md-8 col-md-offset-2">
+      <div className="col-md-12">
```
Elorfin: Completely remove the grid here. This is not needed as it's always full width. Tyou can avoid 2 `div` containers.

plugin/exo/Resources/modules/quiz/overview/overview.jsx
```diff
@@ -137,7 +137,7 @@ const Layout = props =>
       </div>
     </div>
     <div className="row">
-      <div className="col-md-8 col-md-offset-2 text-center">
+      <div className="col-md-12 text-center">
```
Elorfin: Completely remove the grid here. This is not needed as it's always full width. Tyou can avoid 2 `div` containers.

PR #1838, opened 2017-01-24
[exo-v2] Set questions editor player paper and feedback

pitrackster:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no


plugin/exo/Serializer/Question/Type/SetQuestionSerializer.php
```diff
@@ -101,50 +129,160 @@ public function deserialize($data, $setQuestion = null, array $options = [])
...
-        }
 
-        // Serialize odd
-        $solutions->odd = [];
-        foreach ($setQuestion->getLabels() as $label) {
-            $solutions->odd[] = $this->serializeItemSolution($label);
+            if (null === $label) {
+                // Create a new Label
+                $label = new Label();
+                $label->setUuid($itemData->id);
```
Elorfin: Remove this line and...

plugin/exo/Serializer/Question/Type/SetQuestionSerializer.php
```diff
@@ -101,50 +129,160 @@ public function deserialize($data, $setQuestion = null, array $options = [])
...
+            }
+
```
Elorfin: Add this : 
```
// Force client ID if needed
if (!in_array(Transfer::USE_SERVER_IDS, $options)) {
    $label->setUuid($secondSetData->id);
}
```

Note the new block is outside of the `if`.

FYI : this permits to let the server calculates the UUIDs for the case of copy or import (when client IDs already exists in the DB).

plugin/exo/Serializer/Question/Type/SetQuestionSerializer.php
```diff
@@ -101,50 +129,160 @@ public function deserialize($data, $setQuestion = null, array $options = [])
...
+                    unset($proposalsEntities[$entityIndex]);
+                    break;
+                }
+            }
+
+            if (null === $proposal) {
+                // Create a new Proposal
+                $proposal = new Proposal();
+                $proposal->setUuid($setData->id);
```
Elorfin: Same modification than the Labels is needed here to manage UUIDs.

PR #1858, opened 2017-01-27
[exo-v2] Fixes exo access for non-admin

kitan1982:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no

### DIFF
--- a/plugin/exo/Resources/modules/quiz/decorators.js
+++ b/plugin/exo/Resources/modules/quiz/decorators.js
@@ -7,7 +7,7 @@ import {makeId} from './../utils/utils'
 
 // augment normalized quiz data with editor state attributes and default values
 // (can be passed an array of sub-decorators for each item mime type)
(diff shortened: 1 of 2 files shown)


plugin/exo/Resources/modules/quiz/decorators.js
```diff
@@ -36,7 +36,10 @@ export function decorate(state, itemDecorators = {}) {
   })
 }
 
-export function decorateItem(item, subDecorator = item => item) {
+export function decorateItem(item, subDecorator = item => item, apply = true) {
+  if (!apply) {
+    return item
+  }
```
stefk: Instead of passing a parameter to the function to cancel its execution, wouldn't it be simpler/clearer not to call it in the first place?
```js
return applyOnItems ? decorateItem(item, subDecorator) : item
```

PR #1859, opened 2017-01-27
[ExoV2] Graphic question editor

stefk:


plugin/exo/Entity/Content/Image.php
```diff
@@ -12,6 +13,8 @@
  */
 class Image
 {
+    use UuidTrait;
```
Elorfin: uuids should be initialized in the constructor (which is not possible through the trait).

`Uuid:uuid4()->toString()`

plugin/exo/Serializer/Question/Type/GraphicQuestionSerializer.php
```diff
@@ -139,6 +159,9 @@ private function deserializeAreas(GraphicQuestion $graphicQuestion, array $solut
                 $area = new Area();
```
Elorfin: You should choose the ID to use for your new entity : 

```
if (!in_array(Transfer::USE_SERVER_IDS, $options)) {
    $area->setUuid($choiceData->id);
}
```

plugin/exo/Serializer/Question/Type/GraphicQuestionSerializer.php
```diff
@@ -85,10 +90,27 @@ private function serializeImage(GraphicQuestion $graphicQuestion)
...
      * @param GraphicQuestion $graphicQuestion
-     * @param \stdClass       $image
+     * @param \stdClass       $imageData
      */
-    private function deserializeImage(GraphicQuestion $graphicQuestion, \stdClass $image)
+    private function deserializeImage(GraphicQuestion $graphicQuestion, \stdClass $imageData)
     {
+        $typeParts = explode('/', $imageData->type);
+        $image = $graphicQuestion->getImage() ?: new Image();
+        $image->setUuid($imageData->id);
```
Elorfin: You should choose the ID to use for your new entity : 

```
if (!in_array(Transfer::USE_SERVER_IDS, $options)) {
    $image->setUuid($choiceData->id);
}
```

PR #1939, opened 2017-02-03
[core] Webpack typescript TESTING

pitrackster:
#### For now typescript compiler handles only ts(x) files
```
/**
 * Associate js(x) files to typescript compiler.
 */
const ts = () => {
  return {
    test: /\.ts(x?)$/,
    loader: 'ts-loader'
  }
}
```

- Only errors in ts files are handled by the compiler... everything used in a js file is not handled by TS compiler so no type hinting at all
- I can set a type hinting in an ES2015 js file (no compiler complaints) but if I use this kind of method in a ts typehinting is checked
- If there is a TS compiler error then the webpack watcher is stuck... We need to relaunch it to see changes...

#### I was not able to make ts-loader to handle js(x) ie make typescript handle everything and get rid of Babel

If ts-loader is configured as below

```
/**
 * Associate js(x) and ts(x) files to typescript compiler.
 */
const ts = () => {
  return {
    test: /\.t|js(x?)$/,
    exclude: /(node_modules|packages)/,
    loader: 'ts-loader'
  }
}
```

For every module in the platform I've got those types of errors

```
ERROR in ./vendor/claroline/distribution/plugin/result/Resources/modules/results/results.js
Module build failed: Error: Could not find file: '/var/www/html/exo-v2/vendor/claroline/distribution/plugin/result/Resources/modules/results/results.js'.
    at getValidSourceFile (/var/www/html/exo-v2/node_modules/typescript/lib/typescript.js:79271:23)
    at Object.getEmitOutput (/var/www/html/exo-v2/node_modules/typescript/lib/typescript.js:79638:30)
    at getEmit (/var/www/html/exo-v2/node_modules/ts-loader/dist/index.js:99:43)
    at Object.loader (/var/www/html/exo-v2/node_modules/ts-loader/dist/index.js:27:11)
 @ multi claroline-distribution-plugin-result-list
``` 
maybe https://www.typescriptlang.org/docs/handbook/compiler-options.html -> moduleResolution can help but i was not able to find a working configuration

#### TS code coverage

see https://github.com/marcules/karma-remap-istanbul


pitrackster:
I've written a set of files in distribution/plugin/exo/Resources/modules/exo/items/set/tests that are used in set editor.... You can use it to play with ts / js files and see how compiler handles everything (npm run watch)

Ther is also a fake plugin in plugin/awesomeTS... you can use it with npm run webpack to see how the files are compiled

PR #2007, opened 2017-02-12
Update from Weblate.

ClaroBot:


om64:
German and Italian are "clean" for the translation work to be done.

Issue #1994, opened 2017-02-10
[ExoBundle] Impossible to "Start the quiz" once we've already click on it


ptsavdar:
| Q             | A
| ------------- | ---
| Bug report?   | yes
| Version?      | 9.0

The first time I tried to answer a quiz by clicking the "Start the quiz" button, everything went fine and the first page was displayed. However when I refreshed the page and clicked again in "Start the quiz" an error pops up and I'm unable to start the quiz ever since. Here's the error:

```
No route found for "GET /api/exercises/6367c1cf-ed0c-11e6-9e89-00505694365a/attempts": Method Not Allowed (Allow: POST)

405 Method Not Allowed - MethodNotAllowedHttpException
1 linked Exception:
MethodNotAllowedException »

[2/2] MethodNotAllowedHttpException: No route found for "GET /api/exercises/6367c1cf-ed0c-11e6-9e89-00505694365a/attempts": Method Not Allowed (Allow: POST)   +

[1/2] MethodNotAllowedException:    +

Logs   -

Stack Trace (Plain Text)   +
```

I don't know if you need any more info, it so please contact me.

ptsavdar:
Also an error occurs when I'm trying to "Display additional info". It's a Javascript error:

```
Uncaught TypeError: Cannot read property '1' of undefined
    at Parameters (claroline-distribution-plugin-exo-quiz-1c6246f….js:54)
    at StatelessComponent.render (claroline-distribution-plugin-exo-quiz-1c6246f….js:18)
    at ReactCompositeComponentWrapper._renderValidatedComponentWithoutOwnerOrContext (claroline-distribution-plugin-exo-quiz-1c6246f….js:18)
    at ReactCompositeComponentWrapper._renderValidatedComponent (claroline-distribution-plugin-exo-quiz-1c6246f….js:18)
    at ReactCompositeComponentWrapper._updateRenderedComponent (claroline-distribution-plugin-exo-quiz-1c6246f….js:18)
    at ReactCompositeComponentWrapper._performComponentUpdate (claroline-distribution-plugin-exo-quiz-1c6246f….js:18)
    at ReactCompositeComponentWrapper.updateComponent (claroline-distribution-plugin-exo-quiz-1c6246f….js:18)
    at ReactCompositeComponentWrapper.receiveComponent (claroline-distribution-plugin-exo-quiz-1c6246f….js:18)
    at Object.receiveComponent (claroline-distribution-plugin-exo-quiz-1c6246f….js:14)
[...]

PR #2022, opened 2017-02-15
[ExoBundle] fixes quiz migration

Elorfin:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Fixed tickets | #1994 

- Adds a new updater to fix migration of quiz types.
- Fixes answer data migration.

**Note about answer migration :**
I've directly fixes the original updater file because the fix is only here to filter incomplete answer data. 
In the old version when a user didn't answered a question, sometimes incomplete data were stored in the DB.

So if someone has already updated to 9.x and do not hit a bug, he doesn't need this fix.

ping @ptsavdar : this should solve your problems.


ptsavdar:
@Elorfin thanx!

PR #2039, opened 2017-02-20
[UrlBundle] Spoof user-agent header when validating URLs

stefk:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no

URLs are currently validated using curl requests but some websites reject them with 406-like errors if they see the default curl user-agent header. This PR simply adds a real browser u-a header to prevent that behaviour.


stefk:
Should be done on 9.x actually.

PR #2066, opened 2017-03-01
Path date condition

CPASimUSante:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Fixed tickets | https://github.com/claroline/Distribution/issues/1216 

This PR allows to have a blocking condition with date on the Path resource

condition-date


plugin/path/Manager/Condition/StepConditionManager.php
```diff
@@ -77,6 +77,14 @@ public function edit(Step $step, StepCondition $condition, \stdClass $conditionS
         // Clean CriteriaGroups to remove
         $this->cleanCriteriaGroups($condition, $createdGroups, $existingGroups);
 
+        //set the date condition
+        $trimmedAvailableFromDate = trim($conditionStructure->availableFromDate, 'Zz');
+        $trimmedAvailableUntilDate = trim($conditionStructure->availableUntilDate, 'Zz');
```
Elorfin: I don't understand why this 2 lines are needed.

plugin/path/Resources/modules/condition/Partial/edit.html
```diff
@@ -33,6 +33,50 @@ <h4 class="panel-title">
                 {{ 'condition_step_activity_notmanageable'|trans:{}:'path_wizards' }}
             </div>
 
+            <!-- if conditiondate is chosen -->
+            <div class="row form-group">
+                <label class="control-label col-md-3">{{ 'condition_date_name'|trans:{}:'path_wizards' }} :</label>
+                <div class="col-md-9 conditiondate">
```
Elorfin: **coding style :** using `condition-date` would be better.

You should use train case for classes and ids.

Elorfin: If you don't plan to style it, just don't add the class  (I don't see any changes in less files in your PR).

plugin/path/Resources/modules/condition/Partial/edit.html
```diff
@@ -33,6 +33,50 @@ <h4 class="panel-title">
...
+                    <ul>
+                        <li>
+                            {{ 'condition_date_availableFromDate'|trans:{}:'path_wizards' }}
+                            <span class="input-group">
+                                <input class="form-control" type="text" id="conditiondatefrom"
+                                    data-ng-model="conditionEditCtrl.step.condition.availableFromDate"
+                                    uib-datepicker-popup="{{ conditionEditCtrl.filterAvailableFromDate['format'] }}"
+                                    is-open="conditionEditCtrl.filterAvailableFromDate['open']"
```
Elorfin: **coding style :** use `data-*` for angular attributes.

CPASimUSante:
@Elorfin : ahh ok, i understand why now the Docker is broken  :) I missed the mirgation files... I'll do it first thing in the morning.

plugin/path/Manager/Condition/StepConditionManager.php
CPASimUSante: If i remember correctly, it is because the php/symfony date format and JS/datepicker dateformat are not directly usable : there is a 1h delay (= timezone offset) between js/php (if i choose 23/02/2017 in the picker it would save 22/02/2017 23:00 in mysql), and the format is not the same.
Anyways i only made it work using this "hack" (?)

plugin/path/Resources/modules/condition/Partial/edit.html
CPASimUSante: i removed it

CPASimUSante:
New docker : http://pr-2066-1488485317.travis-build.claroline.net/claroline/workspaces/295/open/tool/resource_manager 

Test the 3 paths with user / user

LaurentGruber:
@CPASimUSante tu pourrais résoudre le conflit ? Il y a encore des choses à faire ou on peut merger ensuite ?

PR #2069, opened 2017-03-02
[ExoBundle] Allows to create item objects

kitan1982:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no


main/core/Library/Utilities/FileUtilities.php
```diff
@@ -102,6 +102,50 @@ public function createFile(
...
 
+    public function createFileFromData(
+        $data,
+        $mimeType,
+        $fileName,
+        $extension,
+        $objectClass = null,
+        $objectUuid = null,
+        $objectName = null,
+        $sourceType = null
```
stefk: This api is clearly problematic:

- too much parameters
- redundancies between `mimeType`, `fileName` and `extension`
- should probably do more internally with few parameters
- not documented

main/core/Library/Utilities/FileUtilities.php
```diff
@@ -15,7 +15,7 @@
 use Claroline\CoreBundle\Entity\File\PublicFileUse;
 use Claroline\CoreBundle\Persistence\ObjectManager;
 use JMS\DiExtraBundle\Annotation as DI;
-use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Filesystem\Filesystem as SymfonyFileSystem;
```
kitan1982: I had a confilct when trying to upload a file from tinymce

plugin/exo/Resources/translations/ujm_exo.fr.json
```diff
@@ -17,6 +17,7 @@
     "add_hint": "Ajouter un indice",
```
om64: OK for me. I hadn't seen this. I't be a good idea to add a **Translation** label whenever I have to see something important.

Issue #2153, opened 2017-03-21
[ExoBundle] Banque de questions


LaurentGruber:
| Q             | A
| ------------- | ---
| Bug report?   | yes
| Version?      | master

Quand je vais dans ma banque de question j'ai une page blanche.

Dans la console : Error: Unknown item type text/html

om64:
I had the same problem, but don't have it anymore. That bug seems to be corrected. Close the issue?

LaurentGruber:
non le problème est toujours là

PR #2175, opened 2017-03-23
[ExoBundle] exclude content items from question bank results

pitrackster:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| Fixed tickets | 2153

resolves https://github.com/claroline/Distribution/issues/2153

- save user with the content item created
- exclude from question bank all items whose mimeTypes does not begins with `application`

### DIFF
--- a/plugin/exo/Serializer/Item/ItemSerializer.php
+++ b/plugin/exo/Serializer/Item/ItemSerializer.php
@@ -203,7 +203,7 @@ class ItemSerializer extends AbstractSerializer
                 // Item not exist
                 $question = new Item();
             }
-        }        
+        }
 
         // Sets the creator of the Item if not set
         $creator = $question->getCreator();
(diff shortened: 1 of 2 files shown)


plugin/exo/Serializer/Item/ItemSerializer.php
```diff
@@ -205,6 +205,15 @@ public function deserialize($data, $question = null, array $options = [])
...
 
+        // Sets the creator of the Item if not set
+        $creator = $question->getCreator();
+        if (empty($creator) || !($creator instanceof User)) {
+            $token = $this->tokenStorage->getToken();
+            if (!empty($token) && $token->getUser() instanceof User) {
+                $question->setCreator($token->getUser());
+            }
+        }
+
```
Elorfin: Changing this block position does nothing.

PR #2196, opened 2017-03-29
[ExoBundle] implement touch backend

pitrackster:
| Q             | A
| ------------- | ---
| New feature?  | yes

Try to switch between react drag and drop HTML5Backend / TouchBackend depending on user agent


Elorfin:
Using UserAgent is not a reliable way to detect current device.

The best solution I've found for now is something like described in this blog post : https://bluespot.io/2017/01/03/react-drag-and-drop-with-touch.html

pitrackster:
@Elorfin seems great but not perfect either.... 

See the discussion here https://github.com/react-dnd/react-dnd-html5-backend/pull/50

And also the issue here https://github.com/Neilos/react-dnd-html5-with-touch-backend/issues/1

Elorfin:
Yes it's not perfect. But for now it seems to be the most reliable method to implement it. At least we can reuse some of the concepts (aka extend HTML5 drag-n-drop with touch support).

I was not proposing to use his library. I think we should start implement our own solution has there is no ready to use lib on the web.

We need to be realistic : we will not be able to solve this problem just with adding a new lib in the project.

PR #2225, opened 2017-04-05
Log exporter by group

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no

We should be able to filter logs by groups and export the result as an excel/csv file.
This is a draft for a CLI tool doing it but it would be better to do it at the GUI level.

edit: I guess I can remove the CLI tool now.

@ptsavdar Where is the resource log form used ? It's the only only one I didn't touch. Did I miss something else ?

### DIFF
--- a/main/core/Command/Logs/LogsFetcherCommand.php
+++ b/main/core/Command/Logs/LogsFetcherCommand.php
@@ -23,7 +23,7 @@ class LogsFetcherCommand extends ContainerAwareCommand
         parent::configure();
 
         $this->setName('claroline:logs:fetch')
-            ->setDescription('Send a friend request to a Claroline platform');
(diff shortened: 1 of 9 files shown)


ptsavdar:
@ngodfraind Normally it's used when someone clicks on "View tracking" in a resource's dropdown option list.

LaurentGruber:
@ptsavdar can you review this pr ?

ptsavdar:
@LaurentGruber Yes, i'm checking it in my machine. Just one thing for the moment : The command's description is : "Send a friend request to a Claroline platform" while what it really does is to export logs by group. It is possible that it may create confusion in users.

Issue #2203, opened 2017-03-29
[CoreBundle] Displaying bug when creating new user


Pauooo:
| Q             | A
| ------------- | ---
| Bug report?   | yes
| Version?      | master

Wrong alignment of the roles and organizations check boxes (with FF and Chrome)
![capture du 2017-03-29 15-27-09](https://cloud.githubusercontent.com/assets/15014599/24456819/7b28bdb2-1494-11e7-807b-ac63200eafce.png)

PR #2243, opened 2017-04-10
Fixes after new theme

ptsavdar:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Fixed tickets | #2203 

- Website iframe fix
- Wiki section ul and ol list style fix
- Checkboxes and Radios in forms fix


main/core/Resources/less/layout/components/forms.less
```diff
@@ -54,3 +54,16 @@ form .global-error {
...
+form.form-horizontal {
+    label {
+        > input {
+            &[type=checkbox], &[type=radio] {
+                position: relative;
+                margin-left: 0;
+            }
+        }
+    }
+}
```
Elorfin: Note this is a ugly hack to make it work.
In fact it's broken because form themes do not use DOM / classes expected by bootstrap.

But ok for now.

plugin/website/Resources/views/Website/view.html.twig
```diff
@@ -41,9 +42,10 @@
                             {% elseif currentPage.type == pageTypes.url %}
                                 <iframe width="100%" iframe-height-on-load="vm.contentHeight" src="{{ currentPage.url }}"></iframe>
                             {% elseif currentPage.type == pageTypes.resource %}
-                                <iframe width="100%" iframe-height-on-load="vm.contentHeight" src="{{ path('claro_resource_open', {
+                                <iframe width="100%" iframe-height-on-load="vm.contentHeight" class="claroline-app-iframe" src="{{ path('claro_resource_open', {
```
Elorfin: In fact I was wrong : the class `claroline-app-iframe` is not needed on the iframe tag (it's automatically appended to `<html> / <body>` when `?iframe=1`).

main/core/Resources/less/layout/components/forms.less
ptsavdar: Yes you're right I know. But It was the most "harmless" way to guarantee that checkboxes and radios in most of the forms will work without pbs. If I had added the extra div.checkbox / div.radio needed by bootstrap it may have had collateral issues in some forms.

plugin/website/Resources/views/Website/view.html.twig
ptsavdar: Ok I'm removing it then. Thanks for the review!

PR #2262, opened 2017-04-20
[ExoBundle] re-add editable quiz selector

pitrackster:
| Q             | A
| ------------- | ---
| Bug fix?      | yes

Missing selector was causing an error when user finishes it's attempt.
We need to know if this selector was consciously removed and why before merging.


Elorfin:
It has been removed because the info is know stored in the serialised Resource node. You need to use the selector in core/layout/resource. Also I'm not sure the property you use is still exported in  the quiz

PR #2295, opened 2017-04-26
[ExoBundle] can export papers to csv

pitrackster:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

User can export papers in csv format


plugin/exo/Controller/Api/PaperController.php
```diff
@@ -169,35 +190,32 @@ public function deleteAction(Paper $paper)
...
                     $paper->getEnd() ? $paper->getEnd()->format('Y-m-d H:i:s') : '',
                     $paper->isInterrupted(),
-                    $this->paperManager->calculateScore($paper, $totalScoreOn),
+                    $score !== floor($score) ? number_format($score, 2) : $score,
                 ], ';');
             }
-
             fclose($handle);
         }, 200, [
             'Content-Type' => 'application/force-download',
```
Elorfin: Can you put all of this method inside ExerciseManager please ?

plugin/exo/Controller/Api/PaperController.php
```diff
@@ -169,35 +190,32 @@ public function deleteAction(Paper $paper)
...
-
             /** @var Paper $paper */
             foreach ($papers as $paper) {
                 $structure = json_decode($paper->getStructure());
                 $totalScoreOn = $structure->parameters->totalScoreOn && floatval($structure->parameters->totalScoreOn) > 0 ? floatval($structure->parameters->totalScoreOn) : 20;
+                $user = $paper->getUser();
+                $score = $this->paperManager->calculateScore($paper, $totalScoreOn);
                 fputcsv($handle, [
-                    $paper->getUser()->getFirstName().'-'.$paper->getUser()->getLastName(),
+                    $user ? $user->getFirstName().' - '.$user->getLastName() : $this->translator->trans('anonymous', [], 'platform'),
```
Elorfin: You should also test `$paper->isAnonymised()` to know if you need to export the user name.
Also, in both case I would just return an empty col instead of a translation string.

plugin/exo/Resources/modules/api/middleware.js
```diff
@@ -104,6 +104,8 @@ function getResponseData(response) {
   if (contentType && contentType.indexOf('application/json') !== -1) {
     // Decode JSON
     data = response.json()
+  } else if(contentType && contentType.indexOf('application/force-download') !== -1) {
+    data = response.blob()
```
Elorfin: This is no longer needed.

plugin/exo/Resources/modules/quiz/components/quiz.jsx
```diff
@@ -67,6 +68,16 @@ function customActions(props) {
     action: '#papers'
   })
 
+  // Export results
+  if (props.editable && select.papersAdmin) {
```
Elorfin: `select.papersAdmin` doesn't work. Selectors always require the current store as param : `select.papersAdmin(state)`. You should use the selector in `mapStateToProps` to have access to the current state and use `props` here.

plugin/exo/Resources/translations/ujm_exo.en.json
```diff
@@ -470,7 +470,7 @@
     "questions_export": "Export selected questions",
     "questions_selected": "{1} <b>1</b> question selected|[2,Inf[ <b>%count%</b> questions selected",
     "questions_share": "Share selected questions",
-    "quiz_total_score_on": "Record the questionnaire on",
+    "quiz_total_score_on": "Questionnaire score",
```
Elorfin: Don't we use "Quiz" in other translations ?

PR #2324, opened 2017-04-28
Admin : Edit LTI app

david42:

### DIFF
--- a/plugin/lti/Controller/LtiController.php
+++ b/plugin/lti/Controller/LtiController.php
@@ -45,6 +45,38 @@ class LtiController extends Controller
     }
 
     /**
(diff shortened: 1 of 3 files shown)


plugin/lti/Controller/LtiController.php
```diff
@@ -45,6 +45,38 @@ public function appAction(Request $request)
...
+        $form = $this->createForm(new AppType(), $app);
+        $form->handleRequest($request);
+
+        if ($form->isSubmitted() && $form->isValid()) {
+            $em->flush();
+
+            return $this->redirectToRoute('ujm_admin_lti');
+        }
+
+        return $this->render('UJMLtiBundle:Lti:appEdit.html.twig', array(
```
LaurentGruber: Use [] instead of array

PR #2367, opened 2017-05-10
[CoreBundle] Fixing user managements for groups.

### LINKED ISSUE #2001: [CoreBundle] Organization group managament
| Q             | A
| ------------- | ---
| Bug report?   | yes/no
| Version?      | master/7.0/etc.

I've had a bug report saying that adding a organization to a group didn't allow the managers of the organization to manage the group users.

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Fixed tickets | Fixes #2001


ngodfraind:
I might be wrong on that one and I think I should use the special doctrine expr() conditions. I just realised it now.

PR #2368, opened 2017-05-11
[CoreBundle] Permit workspace user CSV import by first and last name

dtisch:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Fixed tickets | 

- A new hidden configuration option ('workspace_users_csv_import_by_full_name' which defaults to false) enables the possibility to import users into a workspace through CSV only by first and last name if the username column is left blank.
- Finding a user by username (if present) takes precedence over the search by first and last name.
- If two or more users are found haring the same first and last name, the import process will be canceled until the desired user to import is associated with his username.


dtisch:
http://pr-2368-1494489748.travis-build.claroline.net/claroline/

PR #2423, opened 2017-05-22
[CoreBundle] Managers can be added in the workspace administration

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no

Administrators management for workspaces in the workspace administration list.


ngodfraind:
Updating this, don't bother with it yet.

ngodfraind:
@Elorfin 

I still need to fix a test, but what do you think about it ?
It include the serializer / api improvements we discussed before.

main/core/API/Finder/Workspace.php
```diff
@@ -0,0 +1,76 @@
...
+                $qb->andWhere("UPPER(obj.{$filterName}) LIKE :{$filterName}");
+                $qb->setParameter($filterName, '%'.strtoupper($filterValue).'%');
+            } else {
+                // catch boolean
+                if ('true' === $filterValue || 'false' === $filterValue) {
+                    $filterValue = 'true' === $filterValue;
+                }
+
+                $qb->andWhere("obj.{$filterName} = :{$filterName}");
+                $qb->setParameter($filterName, $filterValue);
```
Elorfin: IMO we can put this part of the query in the Finder...

main/core/Resources/modules/layout/modal/index.js
```diff
@@ -5,17 +5,20 @@ import {MessageModal} from './components/message.jsx'
...
 export const MODAL_URL = 'MODAL_URL'
+export const MODAL_USER_PICKER = 'MODAL_USER_PICKER'
 
 const modals = {
   [MODAL_MESSAGE]: MessageModal,
   [MODAL_CONFIRM]: ConfirmModal,
   [MODAL_DELETE_CONFIRM]: DeleteConfirmModal,
-  [MODAL_URL]: UrlModal
+  [MODAL_URL]: UrlModal,
+  [MODAL_USER_PICKER]: UserPickerModal
```
Elorfin: I don't know if we really need to register this modal in all pages. You can register it only for workspace management in the app constructor (see quizzes).

main/core/API/Finder/Workspace.php
ngodfraind: I don't think so. It's specific to the object.

PR #2429, opened 2017-05-26
Workspace manager improvements

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no

- [x] translations
- [x] search by date
- [x] user picker can be opened
- [x] model flag in workspace creation
- [ ] double check requests for other workspace listings (remove models) Widget, public list, history
- [x] Default filter : personnal workspace : false & models : false


main/core/API/Finder.php
```diff
@@ -62,13 +62,22 @@ public function search($class, $page, $limit, array $searches = [])
...
         }, $data);
 
+        $filterObjects = [];
+
+        foreach ($filters as $property => $value) {
+            $filterObject = new \stdClass();
+            $filterObject->value = $value;
+            $filterObject->property = $property;
+            $filterObjects[] = $filterObject;
+        }
```
Elorfin: You can just use the original `$filters` var and avoid this extra loop.

main/core/Serializer/Workspace/WorskpaceSerializer.php
```diff
@@ -75,6 +75,8 @@ public function serialize(Workspace $workspace)
           'uuid' => $workspace->getGuid(),
           'name' => $workspace->getName(),
           'code' => $workspace->getCode(),
+          //moment timestamp
+          'dateCreation' => $workspace->getCreationDate()->getTimeStamp() * 1000,
```
Elorfin: Date should be returned in ISO 8601 format `Y-m-d\TH:i:s`.

main/core/Resources/modules/administration/workspace/components/workspaces.jsx
```diff
@@ -124,7 +124,9 @@ class Workspaces extends Component {
               {name: 'isModel', type: 'boolean', label: t('model')},
               {name: 'isPersonal', type: 'boolean', label: t('personal_workspace'), flags: LIST_PROP_DEFAULT&~LIST_PROP_DISPLAYED},
               {name: 'displayable', type: 'boolean', label: t('displayable_in_workspace_list'), flags: LIST_PROP_DEFAULT&~LIST_PROP_DISPLAYED},
-              {name: 'creationDate', type: 'date', label: t('creation_date')},
+              {name: 'dateCreation', type: 'date', label: t('creation_date'), flags:~LIST_PROP_FILTERABLE},
+              {name: 'createdAfter', type: 'date', label: t('created_after'), flags: LIST_PROP_FILTERABLE&~LIST_PROP_DISPLAYABLE&~LIST_PROP_DISPLAYED},
+              {name: 'createdBefore', type: 'date', label: t('created_before'), flags: LIST_PROP_FILTERABLE&~LIST_PROP_DISPLAYABLE&~LIST_PROP_DISPLAYED},
```
Elorfin: IMO, it would be better to merge `createdAfter` / `createdBefore` and to create a new `date_range` data type with a filter that permits to specify both values.

The `date_range` is common and we will need it for many lists. Better bundle all we can for further uses.

LaurentGruber: On a souvent le cas de je veux tout avant cette date, ou tout après celle là et la combinaison des 2 permets un interval donc je pense que ça répond à plus de cas ...

main/core/API/Finder.php
ngodfraind: This is the format required by the javascript.

Elorfin: `json_encode` will convert the associative array into a JS object. Also note that `$filters` comes from the JS app.

ngodfraind: They're slightly different...
The object is 
``` [{"value":false,"property":"isModel"},{"value":false,"property":"isPersonal"}] ```

PR #2694, opened 2017-07-20
[CoreBundle] Solving the workspace administation perf issue.

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes/no
| BC breaks?    | yes/no

There is an issue with the getUsersByRolesIncludingGroups() because it's used with a pager in WorkspaceAnalyticsController::workspaceManagerActivityEvaluationsShowAction()

### DIFF
--- a/main/core/Controller/WorkspaceAnalyticsController.php
+++ b/main/core/Controller/WorkspaceAnalyticsController.php
@@ -186,7 +186,7 @@ class WorkspaceAnalyticsController extends Controller
             // It only allows to prevent 1 DB request per activity when getting
             // resourceNode linked to activity
             $resourceType = $this->resourceManager->getResourceTypeByName('activity');
-            $resourceNodes = $this->resourceManager
+            $this->resourceManager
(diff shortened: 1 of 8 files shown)


main/core/Controller/WorkspaceAnalyticsController.php
```diff
@@ -186,41 +186,41 @@ public function workspaceActivitiesEvaluationsShowAction(
             // It only allows to prevent 1 DB request per activity when getting
             // resourceNode linked to activity
             $resourceType = $this->resourceManager->getResourceTypeByName('activity');
-            $resourceNodes = $this->resourceManager
+            $this->resourceManager
```
Elorfin: Why ?
If it's not an error, the whole function call should be removed.

ngodfraind: This is a way to fetch the entities so doctrine doesn't do more db calls.

PR #2700, opened 2017-07-25
[CoreBundle] Organization picker

### LINKED ISSUE #1178: pb d'ergonomie - Organisation - création de user
Pour une meilleurs visibilité dans les organisation peut on cascader de la meme façon cote création de compte utilisateur comme c'est cascadé coté gestion des organisation

image

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix? | no
| New feature?  | no
| Fixed tickets | Fixes #1178

- New Symfony form field type "organization"
- New React form field type "organization"
- React treeview
- Default entity finder using doctrine "find" method
- Twig method to access the api searcher
- Seralizer can use a $options array


ngodfraind:
summon @Elorfin 

The treeview is almost finished and you could give me some input imo.

I plan to add event on open/close node and easier css for the expand node button.

main/core/Form/Field/OrganizationPickerType.php
```diff
@@ -0,0 +1,81 @@
...
+use Symfony\Component\Form\FormView;
+use Symfony\Component\OptionsResolver\Options;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * @Service("claroline.form.organization_picker")
+ * @FormType(alias = "organizationpicker")
+ *
+ * Required because the normalizer anonymous function screws up PHPMD
+ * @SuppressWarnings(PHPMD)
```
Elorfin: What is the error thrown ? There may be a better way to handle it than disabling the whole validator.

main/core/Form/Field/OrganizationPickerType.php
```diff
@@ -0,0 +1,81 @@
...
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\Form\FormInterface;
+use Symfony\Component\Form\FormView;
+use Symfony\Component\OptionsResolver\Options;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * @Service("claroline.form.organization_picker")
+ * @FormType(alias = "organizationpicker")
```
Elorfin: CS : you should use `orginazation_picker` instead.

main/core/Resources/less/layout/components/treeview.less
```diff
@@ -0,0 +1,47 @@
+.treeview,
+.treeview *
+{
```
Elorfin: You don't have to fix it, but FYI, `{` should be on the same line than the last selector. `.treeview * {`.

Also note, that the `*` should be avoided (performances impact + unexpected behavior if DOM change).

main/core/Resources/modules/layout/treeview/treeview.jsx
```diff
@@ -0,0 +1,189 @@
...
+    flattened.forEach(el => {
+      if (this.hasChildChecked(el)) {
+        opened.push(el.id)
+      }
+    })
+
+    const cssClasses = this.props.options.cssClasses || {}
+    this.commonCss = cssClasses.common || 'treeview-button-common'
+    this.openCss = cssClasses.open || 'fa fa-chevron-circle-down'
+    this.closeCss = cssClasses.close || 'fa fa-chevron-circle-right'
```
Elorfin: You should add `fa-fw` (in both `openCss`, `closeCss`). It ensures all the icons have the same width and have proper alignment in lists.

main/core/Resources/modules/layout/treeview/treeview.jsx
```diff
@@ -0,0 +1,189 @@
...
+            anchorPrefix={this.props.anchorPrefix}
+            data={el.children}
+            options={this.props.options}
+            onChange={this.props.onChange}
+            onOpenNode={this.props.onOpenNode}
+            onCloseNode={this.props.onCloseNode}
+            render={this.props.render}
+          />),
+          document.getElementById(`${this.props.anchorPrefix}-node-${el.id}`)
+        )}
```
Elorfin: You should not use `ReactDOM.render` as it will create a new React app. You can just return the list of `TreeNode` to render.

main/core/Resources/modules/symfony/form/organization/organization_picker.jsx
```diff
@@ -0,0 +1,52 @@
...
+  constructor(props) {
+    super(props)
+    this.renderers = {}
+  }
+
+  render() {
+    return (
+      <TreeView
+        data={this.props.organizations}
+        renderers={this.renderers}
```
Elorfin: 1. This is not documented in PropTypes.
2. You should define a `defaultProp` in your component and remove l.10 and l.17.

main/core/Resources/views/Form/organization_picker.html.twig
```diff
@@ -0,0 +1,31 @@
...
+              'selected': value,
+              'collapse': false
+            }
+            | json_encode | raw | escape
+          }}"
+        >
+      </div>
+    {% endspaceless %}
+    <!-- dunno why I must do this include here-->
+    <script src="{{ hotAsset('dist/angular_dll.js') }}"></script>
```
Elorfin: Are you sure about this one ?

main/core/Serializer/Organization/OrganizationSerializer.php
```diff
@@ -0,0 +1,75 @@
...
+class OrganizationSerializer
+{
+    private $om;
+
+    /**
+     * ResourceNodeManager constructor.
+     *
+     * @DI\InjectParams({
+     *     "om"                = @DI\Inject("claroline.persistence.object_manager"),
+     *     "container"         = @DI\Inject("service_container")
```
Elorfin: Sensio CS : you should not include the whole container.

main/core/Serializer/Organization/OrganizationSerializer.php
```diff
@@ -0,0 +1,75 @@
+<?php
+
+namespace Claroline\CoreBundle\Serializer\Organization;
```
Elorfin: Can you move it inside `Claroline\CoreBundle\API\Serializer` instead ? see https://github.com/claroline/Distribution/pull/2712.

main/core/Serializer/Organization/OrganizationSerializer.php
Elorfin: Anyway, it's not used in the class.

main/core/Twig/SearcherExtension.php
```diff
@@ -0,0 +1,59 @@
...
+use JMS\DiExtraBundle\Annotation\Inject;
+use JMS\DiExtraBundle\Annotation\InjectParams;
+use JMS\DiExtraBundle\Annotation\Service;
+use JMS\DiExtraBundle\Annotation\Tag;
+
+/**
+ * @Service
+ * @Tag("twig.extension")
+ */
+class SearcherExtension extends \Twig_Extension
```
Elorfin: It would have been better named `FinderExtension` for naming consistency.

main/core/Twig/SearcherExtension.php
```diff
@@ -0,0 +1,59 @@
...
+{
+    protected $container;
+
+    /**
+     * @InjectParams({
+     *     "container" = @Inject("service_container")
```
Elorfin: Sensio CS : just inject the FinderProvider instead.

main/core/API/Finder.php
```diff
@@ -49,7 +49,7 @@ public function getFinder($class)
         }
     }
 
-    public function search($class, $page, $limit, array $searches = [])
+    public function search($class, $page, $limit, array $searches = [], array $serializerOptions = [])
```
Elorfin: Note for later : we should move `$page`, `$limit` params in the `$searches` array (and rename it).
1. `$searches` already contains other params like `filters` and `sortBy`. There is no reason to treat `page` and `limit` in a different way.
2. it will simplify the use of the finder.

For example, we might simplify https://github.com/claroline/Distribution/blob/master/main/core/Controller/API/Workspace/WorkspaceController.php#L226.

```
return $this->finder->search(
          'Claroline\CoreBundle\Entity\Workspace\Workspace',
          $this->container->get('request')->query->all()
        );
```

So we don't have to manage pagination per implementation (Users, WS, etc.)

main/core/Resources/views/Form/organization_picker.html.twig
ngodfraind: Uncaught ReferenceError: angular_dll_1421d34e144c3836f3b2 is not defined

My best guess is that this is were the webpack runtime is included.

Elorfin: Yes, weppack is included in the javascripts juste before the end of `body` (https://github.com/claroline/Distribution/blob/master/main/core/Resources/views/Layout/javascripts.html.twig#L5). So it may not be already available here, but you should have errors on other dlls (like es6)...

ngodfraind: It wants it no matter what (even with no import in the field.jsx file).
This is the only missing dll.

PR #2732, opened 2017-07-31
[CoreBundle] adds react-router-dom and core UI components

Elorfin:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no

Adds react-router-dom as dependency (https://www.npmjs.com/package/react-router-dom)

ping @ngodfraind @kitan1982 

Adds a new `Page` redux container (used in WS management) so this is now longer needed to manually bind modales in implementations. Simplifications can be seen here : https://github.com/claroline/Distribution/pull/2732/files#diff-1cdf53377728baa51a013530940ea96b.

**Sugar synthax to bootstrap a react app.**

Implementation is here : https://github.com/claroline/Distribution/blob/react-router/main/core/Resources/modules/utilities/app/bootstrap.js#L15.

It can be seen in action in the workspace management : https://github.com/claroline/Distribution/blob/react-router/main/core/Resources/modules/administration/workspace/index.js#L22.

**Sugar synthax to bootstrap a routed react app.**

Implementation is here : https://github.com/claroline/Distribution/blob/react-router/main/core/Resources/modules/utilities/app/router.js#L53.

For now, it's not used in the 10.x branch. An example, can be found in theme management (which should be merged soon) https://github.com/claroline/Distribution/blob/elorfin-themes/main/core/Resources/modules/administration/theme/index.js#L18.

### DIFF
--- a/main/core/Resources/modules/administration/workspace/components/workspaces.jsx
+++ b/main/core/Resources/modules/administration/workspace/components/workspaces.jsx
@@ -18,8 +18,13 @@ import {select as paginationSelect} from '#/main/core/layout/pagination/selector
 import {select as listSelect} from '#/main/core/layout/list/selectors'
 import {select} from '#/main/core/administration/workspace/selectors'
 
-import {Page, PageHeader, PageContent} from '#/main/core/layout/page/components/page.jsx'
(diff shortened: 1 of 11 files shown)


ngodfraind:
looks nice

PR #2761, opened 2017-08-09
[ExoBundle] Mandatory questions/answers

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Fixed tickets | #2714 

Also fix an issue introduced by the previous PR (maxPapers === 0 is now working).


ngodfraind:
This list of commit is the one from the previous PR so it should be squashed.

PR #2837, opened 2017-08-29
[ExoBundle] Question duplication

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes/no

This should fix it.
Not sure if it's linked but editing a cloze question type is painfully slow atm.


plugin/exo/Resources/modules/utils/utils.js
```diff
@@ -25,6 +25,49 @@ export function lastId() {
   return lastGeneratedIds[lastGeneratedIds.length - 1]
 }
 
+function isObject(item) {
+  return (typeof item === 'object' && !Array.isArray(item) && item !== null)
+}
```
Elorfin: Use the one from `lodash`.

plugin/exo/Resources/modules/utils/utils.js
```diff
@@ -25,6 +25,49 @@ export function lastId() {
...
+
+function isArray(item) {
+  return Array.isArray(item)
+}
```
Elorfin: Use the one from `lodash`.

ngodfraind:
@Elorfin 

J'ai simplifié le code. Ça fonctionne bien comme ça.

PR #2879, opened 2017-09-06
[InwicastClarolinePluginBundle] Import inwicast bundle in distribution

ptsavdar:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Fixed tickets |


composer.json
```diff
@@ -126,7 +126,8 @@
         "innova/audio-recorder-bundle": "*",
         "innova/media-resource-bundle": "*",
         "ujm/exo-bundle": "*",
-        "icap/lesson-bundle": "*"
+        "icap/lesson-bundle": "*",
+        "icap/inwicast-plugin-bundle": "*"
```
Elorfin: It should be : `icap/inwicast-bundle`

composer.json
```diff
@@ -188,7 +189,8 @@
             "Claroline\\CasBundle\\": "plugin/cas",
             "Icap\\BibliographyBundle\\": "plugin/bibliography",
             "Claroline\\ExternalSynchronizationBundle\\": "plugin/external-user-group-synchronization",
-            "Claroline\\BigBlueButtonBundle\\": "plugin/big-blue-button"
+            "Claroline\\BigBlueButtonBundle\\": "plugin/big-blue-button",
+            "Inwicast\\ClarolinePluginBundle\\": "plugin/inwicast"
```
Elorfin: The correct namespace should be : `Icap\InwicastBundle`.

plugin/inwicast/DependencyInjection/Configuration.php
```diff
@@ -0,0 +1,24 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function getConfigTreeBuilder()
+    {
+        $treeBuilder = new TreeBuilder();
+
+        return $treeBuilder;
+    }
+}
```
Elorfin: If you don't use it, you don't need it.

plugin/inwicast/DependencyInjection/InwicastClarolinePluginExtension.php
```diff
@@ -0,0 +1,28 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function load(array $configs, ContainerBuilder $container)
+    {
+        $configuration = new Configuration();
+        $this->processConfiguration($configuration, $configs);
+
+        $loader = new Loader\YamlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
+        $loader->load('services/services.yml');
```
Elorfin: You should prefer annotations for your services definitions. It's easier to read/update and you can remove this class to simplify your bundle.

plugin/inwicast/Entity/Mediacenter.php
```diff
@@ -0,0 +1,217 @@
...
+
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * Mediacenter.
+ *
+ * @ORM\Table(name="inwicast_plugin_mediacenter")
+ * @ORM\Entity(repositoryClass="Inwicast\ClarolinePluginBundle\Repository\MediacenterRepository")
+ */
+class Mediacenter
```
Elorfin: It should be full camelCase : `MediaCenter`. It's the same for all your other "derived" classes (repositories, exceptions, etc.)

plugin/inwicast/Form/MediacenterType.php
```diff
@@ -0,0 +1,58 @@
...
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolverInterface;
+
+/**
+ * Class MediacenterType.
+ *
+ * @DI\FormType;
+ */
+class MediacenterType extends AbstractType
```
Elorfin: Same here : incorrect camelCase.

plugin/inwicast/Installation/AdditionalInstaller.php
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Inwicast\ClarolinePluginBundle\Installation;
+
+use Claroline\InstallationBundle\Additional\AdditionalInstaller as BaseInstaller;
+
+class AdditionalInstaller extends BaseInstaller
+{
+    public function postUpdate($currentVersion, $targetVersion)
+    {
+        if (version_compare($currentVersion, '1.0.2', '<=')) {
```
Elorfin: Are you sure about the version number ? IMHO, it should be 10.

plugin/inwicast/InwicastClarolinePluginBundle.php
```diff
@@ -0,0 +1,31 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Inwicast\ClarolinePluginBundle;
+
+use Claroline\CoreBundle\Library\DistributionPluginBundle;
+use Claroline\KernelBundle\Bundle\ConfigurationBuilder;
+use Inwicast\ClarolinePluginBundle\Installation\AdditionalInstaller;
+
+class InwicastClarolinePluginBundle extends DistributionPluginBundle
```
Elorfin: It should be `IcapInwicastBundle`.

plugin/inwicast/Repository/MediacenterRepository.php
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * MediacenterRepository.
+ *
+ * This class was generated by the Doctrine ORM. Add your own custom
+ * repository methods below.
+ */
+class MediacenterRepository extends EntityRepository
+{
+}
```
Elorfin: If you don't use it, you don't need it.

plugin/inwicast/Resources/public/css/design.css
```diff
@@ -0,0 +1,79 @@
...
+div.caption .caption-info, div.caption .caption-description{
+    color: #555;
+    font-size: 0.9em;
+}
+
+div.tinymce_title_choose {
+    margin: 10px;
+    font-size: 1.2em;
+    color: #444;
+}
```
Elorfin: - It should be a less file in `your_plugin/Resources/styles`.
- It should be loaded via the theme system.

You have to register it like this : https://github.com/claroline/Distribution/blob/10.x/plugin/exo/assets.json#L12.

And then, your your twig you request it with : https://github.com/claroline/Distribution/blob/10.x/plugin/exo/Resources/views/Exercise/open.html.twig#L6

plugin/inwicast/Resources/views/Inwicast/javascript_layout.html.twig
```diff
@@ -0,0 +1,65 @@
...
+ * plugin.js
+ *
+ * Copyright, Moxiecode Systems AB
+ * Released under LGPL License.
+ *
+ * License: http://www.tinymce.com/license
+ * Contributing: http://www.tinymce.com/contributing
+ */
+
+/*jshint unused:false */
```
Elorfin: You must have really good reason to disable linters. This is not one of ours (we use eslint) but you should remove it.

plugin/inwicast/Installation/Updater/Updater010200.php
```diff
@@ -0,0 +1,92 @@
...
+                }
+            }
+        }
+        if ($index % 200 !== 0) {
+            $this->entityManager->flush();
+            $this->entityManager->clear();
+            $this->log(sprintf('    %d ordered tools added - %s', $index % 200, date('Y/m/d H:i:s')));
+        }
+    }
+}
```
Elorfin: Create/Update of tools is managed by the `core`. So this updater is not needed. If it doesn't work, we should fix the core instead.

composer.json
ptsavdar: The folder created by composer is icap/inwicast-plugin-bundle, so the replace command would be false in your case.

composer.json
ptsavdar: The plugin already existed before, mine was just a fork because I didn't have the edit rights in the original one. I kept the same namespace for all platforms that have already installed the plugin. I thought it wouldn't work otherwise. What do you think?

plugin/inwicast/DependencyInjection/Configuration.php
ptsavdar: Ok removed the file.

plugin/inwicast/Installation/AdditionalInstaller.php
ptsavdar: The plugin is quite old. The updater existed already. It has nothing to do with the distribution's 10 version.

plugin/inwicast/InwicastClarolinePluginBundle.php
ptsavdar: Same here, I didn't want to change the classname for all environments that have the inwicast plugin already installed.

plugin/inwicast/Repository/MediacenterRepository.php
ptsavdar: Yeap. Removed it.

plugin/inwicast/Resources/views/Inwicast/javascript_layout.html.twig
ptsavdar: It's in a twig file anyway. I don't think eslint examines .twig files. However I removed it just in case.

plugin/inwicast/Resources/public/css/design.css
ptsavdar: Great! Thanks about that! Didn't know! I admit to be a bit confused about the theme generation but I'm starting to seeing the picture now. Thanks again!

plugin/inwicast/Installation/Updater/Updater010200.php
ptsavdar: I think that what the updater tries to do is to create the tool, if it doesn't already exists. And for every user to add the tool in the user menu, if it is not already there. I think that it was made for all those platforms that had inwicast already installed, then the tool was added but at the time core couldn't handle the tool creation after the plugin installation. I don't know if we really need it now.

ptsavdar: I think that I'll remove it. For all environments that already have the plugin it's gonna work and for the rest, the plugin install will handle it properly.

ptsavdar:
@Elorfin I've made the changes you requested, could you review my answers as well the new code?
 Thanks again!

LaurentGruber:
@ptsavdar ne faudrait-il pas éviter de squasher la branch pour garder l'historique git ?

ptsavdar:
@LaurentGruber Yeap! Thanks! Just labeled it "no-squash"

composer.json
ngodfraind: I don't think the platform will try to remove the old tables because the migrations of the old plugin simply won't be found (so it'll stay as it is) and the namespace can be changed without side effects. But you have to be careful on that one tho.

composer.json
ngodfraind: @ptsavdar is right imo.

composer.json
ptsavdar: @ngodfraind @Elorfin I changed the namespace as asked, but for all the environments that have already installed the plugin, it tries to reinstall it (since now it is called IcapInwicastBundle), so as expected all the migrations are failing, so I commented them but after the bundle is installed I have inwicast tools doubled. One for InwicastClarolinePluginBundle and one for IcapInwicastBundle. And I still have in the DB the InwicastClarolinePluginBundle. Any ideas how I should proceed? Thanks!

ngodfraind: Mmmh. For the migrations part, I currently have no idea and it's going to be an issue later on (well, it's easy to fix but it's annoying to do).

Maybe check if the old name exists in the database and remove it in a Datafixture files ? it's a bit dirty but it should do the trick.

PR #2896, opened 2017-09-08
[CoreBundle] Fixed use imports in thumbnail creator

ptsavdar:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    |no
| Fixed tickets |

### DIFF
--- a/main/core/Library/Utilities/ThumbnailCreator.php
+++ b/main/core/Library/Utilities/ThumbnailCreator.php
@@ -11,9 +11,9 @@
 
 namespace Claroline\CoreBundle\Library\Utilities;
 
+use JangoBrick\SVG\Nodes\Embedded\SVGImageElement;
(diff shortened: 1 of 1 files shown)


Elorfin:
@ngodfraind has already fixed it by upgrading the SVG dep to the correct version.

Elorfin:
see https://github.com/claroline/Distribution/pull/2878.

ptsavdar:
@Elorfin yeap, you're right. I didn't see that. I was having some errors in my installation and thought that someone mistakenly changed it. In that case we need to remove that also: https://github.com/claroline/Claroline/blob/master/composer.json#L94

Elorfin:
Good catch !
I will update the composer.json. Thanks.

Elorfin:
https://github.com/claroline/Claroline/pull/1242

PR #2918, opened 2017-09-14
[10.x] Claco form refactoring

kitan1982:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no


main/core/Resources/modules/layout/modal/components/base.jsx
```diff
@@ -19,7 +19,7 @@ const BaseModal = props =>
             <span className={classes('modal-icon', props.icon)} />
           }
 
-          {props.title}
+          <div dangerouslySetInnerHTML={{__html: props.title}}></div>
```
Elorfin: Don't put HTML in modal title. If you really need it, at least create a new modal.

Also note you should use the component `HtmlText` in `layout/components` to render HTML from string.

main/core/Resources/modules/layout/modal/components/user-picker.jsx
```diff
@@ -9,7 +9,10 @@ import {UserTypeahead} from '#/main/core/layout/typeahead/users/typeahead.jsx'
 const UserPickerModal = props =>
   <BaseModal {...props}>
     <Modal.Body>
-      <UserTypeahead {...props} />
+      {props.content &&
+        <div dangerouslySetInnerHTML={{__html: props.content}}></div>
+      }
+      <UserTypeahead {...props}/>
```
Elorfin: Same here. This should not appear in base component.

plugin/claco-form/API/Finder/EntryFinder.php
```diff
@@ -0,0 +1,285 @@
...
+    /**
+     * EntryFinder constructor.
+     *
+     * @DI\InjectParams({
+     *     "container" = @DI\Inject("service_container")
+     * })
+     *
+     * @param ContainerInterface $container
+     */
+    public function __construct(ContainerInterface $container)
```
Elorfin: You should only inject the services you need, not the whole container.

plugin/claco-form/API/Finder/EntryFinder.php
```diff
@@ -0,0 +1,285 @@
...
+        $translator = $this->container->get('translator');
+        $currentUser = $tokenStorage->getToken()->getUser();
+
+        $isAnon = $currentUser === 'anon.';
+        $clacoForm = $clacoFormManager->getClacoFormById($extraData['clacoForm']);
+        $canEdit = $clacoFormManager->hasRight($clacoForm, 'EDIT');
+        $isCategoryManager = !$isAnon && $clacoFormManager->isCategoryManager($clacoForm, $currentUser);
+        $searchEnabled = $clacoForm->getSearchEnabled();
+        $sortBy = isset($extraData['sortBy']) ? $extraData['sortBy'] : null;
+        $direction = isset($extraData['direction']) ? $extraData['direction'] : null;
```
Elorfin: This is automatically done by the Finder. You don't have to handle it here. And so you can remove the new param you've added `extraData`.

```
$this->finder->search(
                'Claroline\CoreBundle\Entity\Theme\Theme', [
                    'limit' => 20,
                    'sortBy' => 'name', // <----------------- for DESC, use '-name'
                ]
            )
```

main/core/Resources/modules/layout/modal/components/base.jsx
kitan1982: So I will create a new modal

plugin/claco-form/API/Finder/EntryFinder.php
kitan1982: The service I need generates a cyclic exception but I will try to do something

PR #2926, opened 2017-09-15
[CoreBundle] Fixes enhancements ws login

ptsavdar:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Fixed tickets | #2915 

- Fixes redirection after login (avoid MethodNotAllowedException)
- Do not search soft deleted resources
- Disable resource deletion event dispatch when soft delete enabled
- If user has the right to administrate, show hidden resources


main/core/Listener/AuthenticationSuccessListener.php
```diff
@@ -124,13 +125,18 @@ public function onLoginSuccess(InteractiveLoginEvent $event)
...
-
+        $securityRoute = null;
         $securityUri = $request->getSession()->get('_security.main.target_path');
-        $securityRoute = $securityUri ?
-            $this->router->match(
-                preg_replace("/(app_dev.php\/|app_dev.php\/)/i", '', parse_url($securityUri, PHP_URL_PATH))
-            )['_route'] :
-            null;
+        // Get route name if security Uri present
+        if ($securityUri) {
```
ngodfraind: What do this code section do ?

ptsavdar: It gets the security_target_url (redirected to login if user not connected and page not plublicly open), checks if is an allowed route and redirects user to this route. Since the request that handles the login submit form is a POST request and many of the platform routes are GET routes, I'm handling the method not allowed exception for these cases (e.g. someone, not logged in, opened a wiki that was not public, is redirected to login page, he logs in to platform and is expected to be redirected to the wiki page, the problem is that wiki page)

PR #2931, opened 2017-09-18
Adding a generic OAuth2 resource owner type 

amyotj:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Fixed tickets | 

Some parameters have been optionally added to the ConfigurationType in order to make it compatible with a generic resource owner from HWIOAuthBundle.
It seems that the button on the login page appears transparent as it is not handled by the current CSS styles.


LaurentGruber:
il y a toujours beaucoup trop de fichier modifié et des commits de fichier qui n'ont rien avoir avec ta modif, à priori seul le dernier commit est pertinent...

PR #2950, opened 2017-09-20
CEB

om64:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes/no
| BC breaks?    | yes/no

See attached picture:

- Questions don't appear: replaced by "null".
- weird character before the bullets

ceb


om64:
Should be an issue! I'm writing it.

PR #3036, opened 2017-10-04
[10.x] Announcement task

kitan1982:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no

### DIFF
--- a/main/core/API/Serializer/Task/ScheduledTaskSerializer.php
+++ b/main/core/API/Serializer/Task/ScheduledTaskSerializer.php
@@ -67,7 +67,7 @@ class ScheduledTaskSerializer
             ],
             'users' => array_map(function (User $user) {
                 return $this->userSerializer->serialize($user);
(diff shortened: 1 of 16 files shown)


main/core/API/Serializer/Task/ScheduledTaskSerializer.php
```diff
@@ -67,7 +67,7 @@ public function serialize(ScheduledTask $scheduledTask)
             ],
             'users' => array_map(function (User $user) {
                 return $this->userSerializer->serialize($user);
-            }, $scheduledTask->getUsers()->toArray()),
+            }, $scheduledTask->getUsers()),
```
Elorfin: I'm not sure about this one. Users is an `ArrayCollection` and `array_map` expects a real array.

kitan1982: `getUsers` function already applies `toArray()` on the `ArrayCollection`

Issue #3168, opened 2017-10-26
Supprimer ressource de type annonce


david42:
| Q             | A
| ------------- | ---
| Bug report?   | yes
| Version?      | 10.x

La suppression génère une erreur : 

request.CRITICAL: Uncaught PHP Exception Doctrine\ORM\ORMInvalidArgumentException: "A new entity was found through the relationship 'Claroline\CoreBundle\Entity\Log\Log#resourceNode' that was not configured to cascade persist operations for entity: aaaa2 / annonces bis. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist  this association in the mapping for example @ManyToOne(..,cascade={"persist"})." at /webapps/claco-test/vendor/doctrine/orm/lib/Doctrine/ORM/ORMInvalidArgumentException.php line 92 {"exception":"[object] (Doctrine\\ORM\\ORMInvalidArgumentException(code: 0): A new entity was found through the relationship 'Claroline\\CoreBundle\\Entity\\Log\\Log#resourceNode' that was not configured to cascade persist operations for entity: aaaa2 / annonces bis. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist  this association in the mapping for example @ManyToOne(..,cascade={\"persist\"}). at /webapps/claco-test/vendor/doctrine/orm/lib/Doctrine/ORM/ORMInvalidArgumentException.php:92)"} [

PR #3169, opened 2017-10-26
Fixes resource copy & deletion by allowing to disable logs

kitan1982:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Fixed tickets | #3168


ngodfraind:
C'est la même chose que https://github.com/claroline/Distribution/pull/3165

PR #3183, opened 2017-10-31
[CoreBundle] Store resource node view count

dtisch:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Fixed tickets | 

This PR : 
- stores the number of times a resource is viewed instead of counting rows from the log table.
- adds this information to the serialized resource node when requested through the API

### DIFF
--- a/main/core/API/Serializer/Resource/ResourceNodeSerializer.php
+++ b/main/core/API/Serializer/Resource/ResourceNodeSerializer.php
@@ -111,6 +111,7 @@ class ResourceNodeSerializer
             ],
             'shortcuts' => $this->getShortcuts($resourceNode),
             'breadcrumb' => $this->breadcrumbManager->getBreadcrumb($resourceNode),
+            'views_count' => $resourceNode->getViewsCount(),
         ];
 
         if (!empty($resourceNode->getWorkspace())) {
--- a/main/core/Entity/Resource/ResourceNode.php
+++ b/main/core/Entity/Resource/ResourceNode.php
@@ -306,6 +306,13 @@ class ResourceNode
      */
     protected $accesses = [];
 
+    /**
+     * @var int
+     *
+     * @ORM\Column(nullable=true, type="integer", name="views_count")
+     */
+    protected $viewsCount;
+
     public function __construct()
     {
         $this->guid = Uuid::uuid4()->toString();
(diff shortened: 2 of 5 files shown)


main/core/Migrations/pdo_mysql/Version20171030163115.php
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * Generation date: 2017/10/30 04:31:16
+ */
+class Version20171030163115 extends AbstractMigration
+{
+    public function up(Schema $schema)
+    {
+        $this->addSql('
+            ALTER TABLE claro_resource_node 
+            ADD views_count INT DEFAULT NULL
```
Elorfin: Default should be `0`.

main/core/Entity/Resource/ResourceNode.php
```diff
@@ -985,4 +992,30 @@ public function getAccessCode()
...
+     */
+    public function addView()
+    {
+        if ($this->getViewsCount() === null) {
+            $this->viewsCount = 1;
+        } else {
+            ++$this->viewsCount;
+        }
+
+        return $this;
```
Elorfin: - You should not use getter inside entity implementation.
- Default `viewsCount` with `0` so you can avoid your condition.

main/core/API/Serializer/Resource/ResourceNodeSerializer.php
```diff
@@ -111,6 +111,7 @@ public function serialize(ResourceNode $resourceNode)
             ],
             'shortcuts' => $this->getShortcuts($resourceNode),
             'breadcrumb' => $this->breadcrumbManager->getBreadcrumb($resourceNode),
+            'views_count' => $resourceNode->getViewsCount(),
```
Elorfin: IMO, it shoud go in the `meta` block, and be named `views`.

main/core/Entity/Resource/ResourceNode.php
dtisch: Defaulting viewsCount with 0 breaks, for example, workspace creation if this part of code is not updated. Keeping this test an null as default value ensure that every part of the application dealing with resource node creation continue to function as intended.

main/core/Migrations/pdo_mysql/Version20171030163115.php
dtisch: Defaulting viewsCount with 0 breaks, for example, workspace creation if this part of code is not updated. Keeping this test an null as default value ensures that every part of the application dealing with resource node creation continues to function as intended.

main/core/Entity/Resource/ResourceNode.php
Elorfin: Why does it break ? What is the error ? It shouldn't if all default are correctly set (in prop declaration in the entity + in migration).

main/core/API/Serializer/Resource/ResourceNodeSerializer.php
dtisch: Right!

main/core/Entity/Resource/ResourceNode.php
dtisch: Code doesn't break anymore. My entity was lacking the second definition of default value. Thanks!

PR #3185, opened 2017-10-31
[ExoBundle] Fix multiple choice question duplicate

ptsavdar:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Fixed tickets | 

Fixes multiple choice question duplication bug. User Id was interpreted as string and as a result refreshIds function in utils/utils.js interpreted it as string to replace and replaced every occurence of userId e.g. 2 with the new uuid it generated. As a result, both content and ids containing "2" were replaced by new uuid.


plugin/exo/Serializer/UserSerializer.php
```diff
@@ -46,7 +46,7 @@ public function serialize($user, array $options = [])
     {
         $userData = new \stdClass();
 
-        $userData->id = (string) $user->getId();
+        $userData->id = (int) $user->getId();
```
Elorfin: Why ? `id` is already an `int` we transtype it to `string` only to be compliant with the JSON Schema (which need to support UUID).

ptsavdar: Because it won't work as a string. Take a look at the code I'm mentioning in utils/utils.js https://github.com/claroline/Distribution/blob/10.x/plugin/exo/Resources/modules/utils/utils.js#L49 if user id is string and let's say "2" then in mapNewIds is going to get falsely detected as an id to replace with a new one and then function refreshIds will replace all occurences of number "2" by the new uuid it generated. If user's id was a uuid instead of a simple id then it would work. However, again it will be false as when copying a question we don't want to change the creator's id with a new one. The creator stays the same, the only ids need changing are the question and choice ids.

Elorfin:
I close it because the fix will not be sufficient to fix the problem (and will have some unpredictable and not wanted effects).

I've opened an issue to keep it in mind when we will refactorize the quiz plugin.

PR #3270, opened 2017-11-28
[TagBundle]Tag creation fix

nicolasjacquemet:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Fixed tickets | #3269

fixes tag creation, new tag entities were missing their ids because they were not flushed, and thus following code would fail.


nicolasjacquemet:
Tagging a resource from resource manager fails when the tag is brand new (not already in DB), described in #3272

nicolasjacquemet:
Changed the fix, no longer forces flush instead checks if the tag is scheduled for insertion

PR #3375, opened 2018-01-09
[CoreBundle] Users change password

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no

Include some bugfixes for enable/disable users aswell.

### DIFF
--- a/main/core/Resources/modules/administration/user/user/actions.js
+++ b/main/core/Resources/modules/administration/user/user/actions.js
@@ -3,6 +3,7 @@ import {generateUrl} from '#/main/core/api/router'
 import {API_REQUEST} from '#/main/core/api/actions'
 import {actions as listActions} from '#/main/core/data/list/actions'
 import {actions as formActions} from '#/main/core/data/form/actions'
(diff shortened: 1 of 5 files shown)


main/core/Resources/modules/administration/user/user/actions.js
```diff
@@ -64,23 +64,40 @@ actions.addOrganizations = (id, organizations) => ({
...
+actions.changePassword = (user, plainPassword) => ({
+  [API_REQUEST]: {
+    url: ['apiv2_user_update', {id: user.id}],
+    request: {
+      method: 'PUT',
+      body: JSON.stringify(Object.assign({}, user, {plainPassword}))
+    },
+    success: (data, dispatch) => {
+      dispatch(listActions.invalidateData('users.list'))
+    }
```
Elorfin: Can you put this method in `#/main/core/user` instead ? This will also be used in the profile change password.

ngodfraind: What about the success callback ? (it's doing data invalidation here)

Issue #3409, opened 2018-01-15
Replace `fabpot/php-cs-fixer` dependency


Elorfin:
| Q             | A
| ------------- | ---
| Bug report?   | no
| Version?      | master/10.x

The lib is abandonned and have been replaced by : https://github.com/FriendsOfPHP/PHP-CS-Fixer.

PR #3492, opened 2018-01-31
[CoreBundle] Tos and translations.

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | yes/no

There are many inconsistencies on how the locale/languages are handled currently. 
This PR doesn't solve anything. It just makes the translatable tos works and that's all.

There are a few things that should be done (maybe later, I don't know) to make things work properly.

- [ ] split the locale and the language variables. Or at least do something. As of now, we have issues because sometimes we use 'fr' for the language/locale, but the broswer/lbraries want to user fr_FR or stuff like that. This is very annoying and should be fixed.

- [ ] The current implementation for translated fields is garbage at the API level. It should be refactored but since it already works, I guess that's it's an ok first draft. It's also widly inconsistent because it uses some totaly unrelated entities and doesn't make sense at all (to be fair, it's the way it was designed so I'm just repairing stuff).

- [ ] The GUI should be improved. Currently we need to switch the language with the footer button. It should be directly in the text area.

- [ ] add some tests because it's troublesome

- [x] fix php-cs. This is a pain for me because we use an outdated version (see claroline/Distribution#3409)


main/core/Resources/modules/data/types/translated/utils.js
```diff
@@ -0,0 +1,15 @@
...
+
+    return tmp.textContent || tmp.innerText || ''
+  }
+
+  return ''
+}
+
+export {
+  getPlainText
+}
```
Elorfin: Use : import {getPlainText} from '#/main/core/data/types/html/utils'

Elorfin: And it's not used. Just remove it please.

PR #3598, opened 2018-02-20
[PathBundle] implements new plugin version

Elorfin:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no


plugin/path/Serializer/PathSerializer.php
```diff
@@ -125,25 +127,35 @@ private function serializeStep(Step $step)
...
 
         foreach ($stepsData as $stepData) {
-            $step = $this->deserializeStep($stepData, null, ['path' => $path, 'order' => $order]);
+            $step = $this->deserializeStep($stepData, null, $newStepsUuids, ['path' => $path, 'order' => $order]);
             $path->addStep($step);
             ++$order;
         }
+        foreach ($oldSteps as $step) {
+            if (!in_array($step->getUuid(), $newStepsUuids)) {
+                $this->om->remove($step);
```
Elorfin: Should be managed with `orphanRemoval` and cascades on the relationship.

PR #3609, opened 2018-02-21
[CoreBundle] Workspace slug opening url.

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | yes/no


main/core/Library/Installation/AdditionalInstaller.php
```diff
@@ -53,7 +53,7 @@ public function preUpdate($currentVersion, $targetVersion)
             $updater = new Updater\Updater110000($this->container);
             $updater->lnPictureDirectory();
             $updater->lnPackageDirectory();
-        } catch (\Exception $e) {
+        } catch (\Exceuuidption $e) {
```
Elorfin: Typo

main/core/Library/Installation/Updater/Updater110300.php
```diff
@@ -0,0 +1,53 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: panos
+ * Date: 9/12/17
+ * Time: 3:30 PM.
+ */
```
Elorfin: This doc block should be removed.

main/core/Library/Installation/Updater/Updater110300.php
```diff
@@ -0,0 +1,53 @@
...
+            $workspace->setSlug(null);
+            $this->om->persist($workspace);
+
+            if (0 === $i % self::BATCH_SIZE) {
+                $this->om->flush();
+            }
+        }
+
+        $this->om->flush();
+    }
```
Elorfin: What's the purpose of the command ? Why would you want to unset all WS slugs ? (Also it can easily be done with only one SQL query).

ngodfraind: According to the doc, if a slug is set to null, it's regenerated. Because there is no slug at the beginning, this is our way of generating them all (even if it's counter intuitive).

PR #3673, opened 2018-03-05
[CoreBundle] ui dependencies, webpack, karma

Elorfin:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no

It will require a PR on `claroline/Claroline` (https://github.com/claroline/Claroline/pull/1379).

- Cleans and fixes Webpack configurations.

- Restores karma tests suite. Karma is runable but still broken. It will require more work to fix existing tests.

- Removes custom installer for deps without `package.json` (ping @ngodfraind). The installer clones source in `node_modules` which makes most of `npm` commands fail (for example, we are no longer able to run `npm shrinkwrap` which is required by jenkins to update deps). Also note there is only one dependency managed like this.

- Removes `tag-it` lib. It's not used anymore.


Elorfin:
TODO : use webpack encore `configureDefinePlugin` instead of custom plugin registration.

PR #3680, opened 2018-03-06
[CoreBundle] exposes configuration & current user in JS

Elorfin:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no

ping @ngodfraind @kitan1982 

- Makes the current user available everywhere in JS apps : 

```js
import {currentUser} from '#/main/core/user/current'

// it will be null if no user logged
const me = currentUser()
```

**Attention this is not managed by redux, so it will not change at runtime for now.**

- Exposes relevant platform configuration to UI (like current locale, name, see `PlatformSerializer` to know what it contains). 

```js
import {platformConfig} from '#/main/core/platform'
// import {platformConfig} from '#/main/core/platform/config'

const wholeConf = platformConfig()

// I take lodash selectors as param
const currentLocale = platformConfig('locale.current')
```

**Attention this is not managed by redux, so it will not change at runtime.**
**Attention the configuration should be kept as short as possible and not contain dynamic data.**


main/core/Resources/views/base.html.twig
```diff
@@ -4,7 +4,7 @@
 
 {% set isRenderedInIframe = app.request.query.get('iframe') %}
 
-<html lang="{{ app.request.locale|default('fr') }}"
+<html lang="{{ app.request.locale|default('en') }}"
```
LaurentGruber: je pense que le default doit rester fr, pour le moment, c'est la langue principal du projet ...

Elorfin: Je peux le changer, mais en vrai on tombe jamais dans le cas. Il y a toujours une locale dans `request`. Le code pourrait simplement être `<html lang="{{ app.request.locale }}"`.

LaurentGruber: ok, np dans ce cas

PR #3684, opened 2018-03-06
[CoreBundle] Exposes current user in JS as JSON object

dtisch:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Fixed tickets |

### DIFF
--- a/main/core/Resources/modules/user/current.js
+++ b/main/core/Resources/modules/user/current.js
@@ -19,7 +19,7 @@ function load() {
 
   invariant(userEl, 'Can not find current user.')
 
-  user = userEl.dataset.user || null
+  user = JSON.parse(userEl.dataset.user) || null
   userLoaded = true
 }
 


main/core/Resources/modules/user/current.js
```diff
@@ -19,7 +19,7 @@ function load() {
 
   invariant(userEl, 'Can not find current user.')
 
-  user = userEl.dataset.user || null
+  user = JSON.parse(userEl.dataset.user) || null
```
Elorfin: You need an additional check. `JSON.parse` fails on empty strings (which is the case for anonymous).

`user = userEl.dataset.user && 0 !== userEl.dataset.user.length ? JSON.parse(userEl.dataset.user) : null`

dtisch: When nobody is logged in, data-user doesn't contain an empty string but 'null'
image
JSON.parse doesn't fail on this value and returns null, which is what we need. Do we really an additional check ? We could rather simplify the the line 22 into :
`user = JSON.parse(userEl.dataset.user)`
and obtain the needed result.

PR #3772, opened 2018-03-19
Text resource refactoring

kitan1982:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no


main/core/Listener/Resource/TextListener.php
```diff
@@ -154,16 +153,21 @@ public function onCopy(CopyResourceEvent $event)
     public function onOpen(OpenResourceEvent $event)
     {
         $text = $event->getResource();
-        $collection = new ResourceCollection([$text->getResourceNode()]);
-        $isGranted = $this->container->get('security.authorization_checker')->isGranted('EDIT', $collection);
-        $revisionRepo = $this->container->get('doctrine.orm.entity_manager')
-            ->getRepository('ClarolineCoreBundle:Resource\Revision');
+        $user = $this->container->get('security.token_storage')->getToken()->getUser();
+
+        $followerResource = $user !== 'anon.' ?
```
Elorfin: IMO, you should inject it inside the serialized ResourceNode through this : https://github.com/claroline/Distribution/blob/11.x/main/core/API/Serializer/Resource/ResourceNodeSerializer.php#L142.

You can see it used here : https://github.com/claroline/Distribution/blob/11.x/plugin/social-media/Listener/ApiListener.php.

This will simplify later use (you don't have to pass additional data to Resource components) and will be available for all resource types without the need to add this block of code to all `open` listeners.

PR #3803, opened 2018-03-21
[CoreBundle] Some security fixes (role/orga)

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes/no
| BC breaks?    | yes/no
| Fixed tickets | #3774  

- [x] properly display role list
- [x] voter verification


ngodfraind:
We should write voter tests eventually.

PR #3835, opened 2018-03-27
[CoreBundle] Workspace parameters react

ngodfraind:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | yes/no
| Fixed tickets | #3823


main/core/API/Serializer/WorkspaceSerializer.php
```diff
@@ -72,7 +73,7 @@ public function serialize(Workspace $workspace, array $options = [])
         if (!in_array(Options::SERIALIZE_MINIMAL, $options)) {
             $serializer = $this->container->get('claroline.api.serializer');
             $serialized = array_merge($serialized, [
-                'poster' => '', // todo : add as Workspace prop
+                'thumbnail' => $workspace->getThumbnail() ? $this->container->get('claroline.serializer.public_file')->serialize($workspace->getThumbnail()) : null,
```
Elorfin: Don't use container, inject the serializer (if possible avoid the provider too).

main/core/API/Serializer/WorkspaceSerializer.php
```diff
@@ -119,6 +121,11 @@ private function getMeta(Workspace $workspace)
             'description' => $workspace->getDescription(),
             'created' => $workspace->getCreated()->format('Y-m-d\TH:i:s'),
             'creator' => $workspace->getCreator() ? $this->userSerializer->serialize($workspace->getCreator(), [Options::SERIALIZE_MINIMAL]) : null,
+            'usedStorage' => $this->container->get('claroline.utilities.misc')
```
Elorfin: Don't use container.

main/core/API/Serializer/WorkspaceSerializer.php
```diff
@@ -134,6 +141,22 @@ private function getDisplay(Workspace $workspace)
         ];
     }
 
+    private function getOptions(Workspace $workspace)
+    {
+        $options = $this->workspaceManager->getWorkspaceOptions($workspace)->getDetails();
+
+        if ($options['workspace_opening_resource']) {
+            $resource = $this->container->get('claroline.api.serializer')->deserialize(
```
Elorfin: Don't use container, inject the serializer (if possible avoid the provider too).

main/core/API/Serializer/WorkspaceSerializer.php
```diff
@@ -134,6 +141,22 @@ private function getDisplay(Workspace $workspace)
...
+              'Claroline\CoreBundle\Entity\Resource\ResourceNode',
+               ['id' => $options['workspace_opening_resource']]
+            );
+
+            $options['opened_resource'] = $this->container->get('claroline.api.serializer')->serialize($resource);
```
Elorfin: We use `camelCase` in serialized objects.

main/core/API/Serializer/WorkspaceSerializer.php
```diff
@@ -176,16 +199,26 @@ private function getRegistration(Workspace $workspace)
      */
     public function deserialize(array $data, Workspace $workspace, array $options = [])
     {
+        // remove this later (with the Trait)
+        //$this->genericSerializer->deserialize($data, $workspace, $options);
+
+        if (isset($data['thumbnail']) && isset($data['thumbnail']['id'])) {
+            $thumbnail = $this->container->get('claroline.api.serializer')->deserialize(
```
Elorfin: Don't use container, inject the serializer (if possible avoid the provider too).

main/core/API/Serializer/Facet/PanelFacetSerializer.php
```diff
@@ -48,9 +48,11 @@ public function serialize(PanelFacet $panel, array $options = [])
             'title' => $panel->getName(),
             'position' => $panel->getPosition(),
             'roles' => $this->serializeRoles($panel->getPanelFacetsRole()->toArray()),
+            //deprecated
             'display' => [
                 'collapsed' => $panel->isDefaultCollapsed(),
             ],
+            'defaultOpened' => true,
```
Elorfin: Why do you replace the previous param by a hardcoded one ?

main/core/API/Serializer/WorkspaceSerializer.php
```diff
@@ -176,16 +199,26 @@ private function getRegistration(Workspace $workspace)
      */
     public function deserialize(array $data, Workspace $workspace, array $options = [])
     {
+        // remove this later (with the Trait)
+        //$this->genericSerializer->deserialize($data, $workspace, $options);
```
Elorfin: No commented code.

main/core/API/Serializer/WorkspaceSerializer.php
```diff
@@ -194,6 +227,19 @@ public function deserialize(array $data, Workspace $workspace, array $options =
         $this->sipe('registration.selfRegistration', 'setSelfRegistration', $data, $workspace);
         $this->sipe('registration.selfUnregistration', 'setSelfUnregistration', $data, $workspace);
 
+        if (isset($data['restrictions']['accessibleFrom'])) {
+            $workspace->setStartDate(DateNormalizer::denormalize($data['restrictions']['accessibleFrom']));
+        }
+
+        if (isset($data['restrictions']['accessibleUntil'])) {
+            $workspace->setEndDate(DateNormalizer::denormalize($data['restrictions']['accessibleUntil']));
+        }
```
Elorfin: `$data['restrictions']` may not exist.

main/core/Controller/APINew/ResourceNodeController.php
```diff
@@ -41,6 +60,31 @@ public function portalSearchAction(Request $request)
...
     }
 
+    /**
+     * @Route("/resourcespicker", name="apiv2_resources_picker", options={ "method_prefix" = false })
+     *
+     * @param Request $request
+     *
+     * @return array
+     */
+    public function resourcesPickerAction(Request $request)
```
Elorfin: I think `ResourceManager` and `ResourcePicker` will be filled with same API endpoint, so the dedicated naming may not be required.

main/core/Controller/APINew/ResourceNodeController.php
```diff
@@ -41,6 +60,31 @@ public function portalSearchAction(Request $request)
...
+    {
+        $user = $this->tokenStorage->getToken()->getUser();
+        if ('anon.' === $user) {
+            throw new AccessDeniedException();
+        }
+        $options = $request->query->all();
+        $options['hiddenFilters']['active'] = true;
+        $options['hiddenFilters']['resourceTypeBlacklist'] = ['directory', 'activity'];
```
Elorfin: - `activity` will be deleted in the near future so there is no need to hardcode it.
- you should not blacklist directories because we will want to be able to open them in the next version of the picker. IMO you should filter it with a regular list filter for now, so we don't have to change api later.

main/core/Manager/WorkspaceManager.php
```diff
@@ -1114,6 +1114,9 @@ public function getWorkspaceOptions(Workspace $workspace)
             $details = [
                 'hide_tools_menu' => false,
                 'background_color' => null,
+                'hide_breadcrumb' => false,
+                'use_workspace_opening_resource' => false,
+                'workspace_opening_resource' => null,
```
Elorfin: `use_workspace_opening_resource` is redundant : you just have to check `null !== workspace_opening_resource` if you want to know if it's enabled.

main/core/Resources/modules/administration/workspace/workspace/components/workspace-card.jsx
```diff
@@ -2,9 +2,12 @@ import React from 'react'
 
 import {t} from '#/main/core/translation'
 import {displayDate} from '#/main/core/scaffolding/date'
+import {Thumbnail} from '#/main/core/workspace/components/thumbnail'
+import {asset} from '#/main/core/scaffolding/asset'
 
 const WorkspaceCard = (row) => ({
-  icon: 'fa fa-book',
+  poster: asset(row.thumbnail.url),
+  //icon: <img src={asset(row.thumbnail.url)} />,
```
Elorfin: should be uncommented and `icon: 'fa fa-book'`

main/core/Resources/modules/data/list/selectors.js
```diff
@@ -1,7 +1,10 @@
 import get from 'lodash/get'
+import {createSelector} from 'reselect'
 
 // retrieves a list instance in the store
-const list = (state, listName) => get(state, listName)
+const list = (state, listName) => {
+  return get(state, listName)
+}
```
Elorfin: Restore short syntax.

main/core/Resources/modules/workspace/components/thumbnail.jsx
```diff
@@ -0,0 +1,32 @@
...
+  alt: T.bool
+}
+
+Thumbnail.defaultProps = {
+  alt: true
+}
+
+export {
+  Thumbnail
+}
```
Elorfin: Nope.

I don't know exactly where you use it but : 
- Thumbnail !== Avatar : this is not the same thing (thumbnail is juste the poster in small), you can not reuse what was done for Avatar.
- This kind of things is managed by generic components.

main/core/Resources/modules/workspace/parameters/components/tool.jsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react'
+
+import {trans} from '#/main/core/translation'
+import {TabbedPageContainer} from '#/main/core/layout/tabs'
+
+import {ParametersActions, ParametersTab} from '#/main/core/workspace/parameters/parameters/parameters.jsx'
+import {DisplayActions, DisplayTab} from '#/main/core/workspace/parameters/parameters/display.jsx'
+
+const Parameters = () =>
+  <TabbedPageContainer
```
Elorfin: This should not be a MultiPage app.
All Workspace parameters can be rendered in only one FormContainer with the correct field groupings.
There are not enough parameters in each section to justify the complexity (code + user experience) of a Multi tool here.

Elorfin: And this is the same FormContainer that should be reused in admin, so you don't have to open the WS to configure all the options.

main/core/API/Serializer/WorkspaceSerializer.php
Elorfin: Also note, the `thumbnail` will not replace the `poster`, this is just the small version of the `poster` for embed and list displays.

main/core/Manager/WorkspaceManager.php
ngodfraind: I know, but this is for backward comp purpose. I didn't rewrite everything yet.

main/core/Resources/modules/workspace/parameters/components/tool.jsx
ngodfraind: There will be at some point because there is the tool/resource right management that should move here later.

Elorfin: IMO it's not a problem to put it in a form panel. Rights tool seems similar to the one for resources and it works in a modal.

main/core/API/Serializer/Facet/PanelFacetSerializer.php
ngodfraind: There is no way to currently allow to  (un)collapse (or I didn't found it).
The new param must be defaultOpened to match the props of the `section` param of the gui.

PR #3836, opened 2018-03-27
[TextPlayerBundle] TextPlayer bundle refactoring

ptsavdar:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes

Passage en React + Ouverture des fichiers .html

### DIFF
--- a/plugin/text-player/Resources/modules/resources/text-file/components/resource.jsx
+++ b/plugin/text-player/Resources/modules/resources/text-file/components/resource.jsx
@@ -4,11 +4,11 @@ import {connect} from 'react-redux'
 
 import {ResourcePageContainer} from '#/main/core/resource/containers/page.jsx'
 import {HtmlText} from '#/main/core/layout/components/html-text'
+import {getPlainText} from '#/main/core/data/types/html/utils'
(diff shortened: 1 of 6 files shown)


plugin/text-player/Resources/modules/resources/text-file/components/resource.jsx
```diff
@@ -0,0 +1,28 @@
...
+import {PropTypes as T} from 'prop-types'
+import {connect} from 'react-redux'
+
+import {ResourcePageContainer} from '#/main/core/resource/containers/page.jsx'
+import {HtmlText} from '#/main/core/layout/components/html-text'
+
+const Resource = props =>
+  <ResourcePageContainer>
+    { props.isHtml && <HtmlText>{props.content}</HtmlText>}
+    { !props.isHtml && <div>{props.content}</div>}
```
Elorfin: IMO, you can :
 
```
<HtmlText>{props.content}</HtmlText>
```

In all case, you avoid the extra param and simplify the code. Also there is a class on HtmlText that justify text (it's more beautiful).

PR #3870, opened 2018-04-04
[ClacoForm] Refactors ClacoForm resource with react components from core

kitan1982:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no


Elorfin:
You also need to make your custom CascadeSelect a `data/types` and add `meta.creatable : true`. You will be able to add it to your `fields` list and handle it with the standard `FormContainer`.

plugin/claco-form/Resources/modules/resources/claco-form/player/entry/components/form-field.jsx
```diff
@@ -13,20 +14,33 @@ import {EmailGroup} from '#/main/core/layout/form/components/group/email-group.j
 import {DateGroup} from '#/main/core/layout/form/components/group/date-group.jsx'
 import {FileGroup} from '#/main/core/layout/form/components/group/file-group.jsx'
 
-// deprecated
-// only used by claco-form
-// todo : rewrite claco-form to use form.jsx
-export const FormField = props => {
+const FormField = props => {
```
Elorfin: This component must be deleted.

kitan1982: It is still in use in ClacoForm when a template is used.
However as it is only used by ClacoForm, it has been move to ClacoForm bundle.

Elorfin: Nope. If you need it, you must use : https://github.com/claroline/Distribution/blob/11.x/main/core/Resources/modules/data/form/components/field.jsx

kitan1982: Ok. It will be switched to that component.

PR #4055, opened 2018-05-03
[ClacoForm][NewMigrations] Fields values & categories association

kitan1982:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no

- Fixes choice field creation
- Small fixes
- Allows to associate values of fields to categories
- Uses a select input for search on country fields


plugin/claco-form/Migrations/pdo_mysql/Version20180502092627.php
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Auto-generated migration based on mapping information: modify it with caution.
+ *
+ * Generation date: 2018/05/02 09:26:29
+ */
+class Version20180502092627 extends AbstractMigration
+{
+    public function up(Schema $schema)
+    {
+        $this->addSql("
```
ngodfraind: Travis passera pas :p

kitan1982: Si, le bloc contient du texte entre quote :
ADD array_value LONGTEXT DEFAULT NULL COMMENT '(DC2Type:json_array)',

PR #4116, opened 2018-05-15
Fixes HTML field when using FormField with a name containing a dot

kitan1982:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no

### DIFF
--- a/main/core/Resources/modules/data/form/components/field.jsx
+++ b/main/core/Resources/modules/data/form/components/field.jsx
@@ -27,7 +27,7 @@ class FormField extends Component {
       )
     } else {
       return React.createElement(typeDef.components.form, merge({}, this.props.options, {
-        id: this.props.name,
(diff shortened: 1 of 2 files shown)


main/core/Resources/modules/data/form/components/field.jsx
```diff
@@ -27,7 +27,7 @@ class FormField extends Component {
       )
     } else {
       return React.createElement(typeDef.components.form, merge({}, this.props.options, {
-        id: this.props.name,
+        id: this.props.name.replace(/\./g, '-'),
```
Elorfin: You could write : `.replace('.', '-')`

kitan1982: Yes but in that case it will only replace the first occurence of '.'

PR #4345, opened 2018-06-28
[AppBundle] Fix Summary component

dtisch:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Fixed tickets | 

Prevent summary from crashing React application when used with a tree having a depth greater than 1.


main/app/Resources/modules/content/summary/components/summary.jsx
```diff
@@ -113,7 +113,7 @@ class SummaryLink extends Component {
           <ul className="step-children">
             {this.props.children.map(child =>
               <SummaryLink
-                key={toKey(child)}
+                key={toKey(child.target)}
```
Elorfin: You can not use it because it's a specific prop of some action types (`link` and `url` for now). So it may not be defined. You must use `label` instead which is required for all actions.

PR #4458, opened 2018-07-20
[NewBundle] Authentication bundle

ptsavdar:
Moved CasBundle and OauthBundle inside this bundle.

LdapBundle and ExternalUserGroupBundle are to be added.


LaurentGruber:
To do : 
- [ ] move  ldap to AuthenticationBundle
- [ ] move external-user-group to AuthenticationBundle

PR #4504, opened 2018-08-08
Adds a function to strip tags from html in front-end

kitan1982:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Fixed tickets | #4247


main/app/Resources/modules/utils/utils.js
```diff
@@ -0,0 +1,41 @@
...
+    processedHtml = processedHtml.replace(
+      /<(video|audio)([^>]+src=[\'"]([^\'"]+)[\'"])*[^\/>]*\/?>([\s\S]*)<\/\1>/i,
+      (matches, tag, src, srcContent, srcs) => mediaSrcExtractor(tag, srcContent, srcs)
+    )
+  }
+  let element = document.createElement('div');
+  element.innerHTML = processedHtml;
+
+  return element.textContent
+}
```
Elorfin: There already is a function for that in `import {getPlainText} from '#/main/app/data/html/utils'` (and there should be other decprecated implementations in exo bundle).
Enhance it if you want, but donc create a new function.

Also note that an `utils/utils` will never be allowed in core and apps. Utils must be classified by concern (and so be located in the correct module).

plugin/exo/Resources/modules/bank/components/questions.jsx
```diff
@@ -62,7 +60,15 @@ const QuestionsPage = props =>
...
             label: tex('question'),
             type: 'html',
-            render: (rowData) => rowData.title || rowData.content.substr(0, 50),
+            render: (rowData) => {
+              if (rowData.title) {
+                return rowData.title
+              } else {
+                const content = stripTagsFromHtml(rowData.content)
+
+                return 50 < content.length ? `${content.substr(0, 50)}...` : content
```
Elorfin: The HTML cell already strip tags if you don't pass `options: {trust: true}`. I think you can add the substr in the cell implementation too (in table we never want to display full text string as it will break the display) and avoid a custom `render`.

PR #4614, opened 2018-08-31
Show resource header

Pauooo:
| Q             | A
| ------------- | ---

| New feature?  | yes


plugin/path/Serializer/PathSerializer.php
```diff
@@ -153,6 +153,7 @@ private function serializeStep(Step $step)
             'description' => $step->getDescription(),
             'poster' => $poster,
             'primaryResource' => $step->getResource() ? $this->resourceNodeSerializer->serialize($step->getResource()) : null,
+            'showResourceHeader' => $step->isShowResourceHeader(),
```
Elorfin: Should be `getShowResourceHeader`

PR #4627, opened 2018-09-03
resource picker current

Pauooo:
| Q             | A
| ------------- | ---
makes that path resourcePicker opens on the WS resource manager


plugin/path/Resources/modules/resources/path/editor/components/step-form.jsx
```diff
@@ -150,7 +150,10 @@ const StepForm = props =>
             label: trans('resource'),
             options: {
               embedded: true,
-              showHeader: true
+              showHeader: true,
+              picker: {
+                current : props.resourceParent
```
Elorfin: Missing prop-types.

PR #5032, opened 2018-10-19
[!!! NEW MIGRATIONS !!!] Refactors user display configuration of desktop tools

kitan1982:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no


main/core/API/Serializer/User/RoleSerializer.php
```diff
@@ -208,6 +213,55 @@ private function serializeTools(Role $role, $workspaceId)
...
+            'dashboard',
+            'formalibre_bulletin_tool',
+            'formalibre_presence_tool',
+            'formalibre_reservation_agenda',
+            'formalibre_support_tool',
+            'home',
+            'inwicast_portal',
+            'my-learning-objectives',
+            'my_portfolios',
+        ];
```
Elorfin: You should maintain this list elsewhere : 

- You redeclare it in LayoutController.
- This is temporary, so it will be easier to find and remove if there is only one definition.

You can create a final class like for API options and put it in. It will be easier to do a search/replace.

main/core/Controller/LayoutController.php
```diff
@@ -160,23 +160,67 @@ public function topBarAction(Workspace $workspace = null, Request $request)
...
+        ];
+        // List of displayable tools in user menu
+        $userToolsWhiteList = [
+            'agenda_',
+            'all_my_badges',
+            'my_contacts',
+            'my_portfolios',
+            'my-learning-objectives',
+            'resource_manager',
+        ];
```
Elorfin: You create an adherence to all plugins by doing it, and you will need to update this for each new tool.

You must store this information in the declaration of the tool in plugin `config.yml`

PR #5056, opened 2018-10-24
Home refactoring

kitan1982:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no

To activate new home page the following option has to be added in the platform_options.yml file :
`home_redirection_type: new`


main/core/Resources/config/config.yml
```diff
@@ -192,18 +193,18 @@ plugin:
...
         - name: resources
           type: list
-          context: [workspace, desktop]
+          context: [workspace, desktop, home]
           tags: [resources]
 
         - name: workspaces
           type: list
-          context: [desktop]
+          context: [desktop, home, home]
```
Elorfin: duplicate key.

main/core/Resources/config/config.yml
```diff
@@ -178,11 +178,12 @@ plugin:
     widgets:
         - name: simple
           class: Claroline\CoreBundle\Entity\Widget\Type\SimpleWidget
-          context: [workspace, desktop]
+          context: [workspace, desktop, home]
           tags: [content]
 
         - name: list
           class: Claroline\CoreBundle\Entity\Widget\Type\ListWidget
+          context: [workspace, desktop, home]
```
Elorfin: I don't think this is required or used (if I remember correctly it will check the context of the DataSource for this cases).

PR #5358, opened 2019-01-10
[TagBundle/ForumBundle] Tag replace forceFlush + forum creator serial…

ngodfraind:
…ization.

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | yes/no
| Fixed issues | #5355


main/core/API/Serializer/MessageSerializer.php
```diff
@@ -118,7 +122,7 @@ public function deserialize($data, AbstractMessage $message, array $options = []
 
             if (isset($data['meta']['creator'])) {
                 $message->setAuthor($data['meta']['creator']['name']);
-                $creator = $this->userSerializer->deserialize($data['meta']['creator']);
+                $creator = $this->serializer->deserialize(User::class, $data['meta']['creator']);
```
Elorfin: Why ?

ngodfraind: The userSerializer doesn't fetch the object if he already exists. So we end up with duplicates and database errors.

Elorfin: OK. Remove `userSerializer` injection as you don't need it anymore.

Issue #3872, opened 2018-04-04
[ExoBundle] implements new version


Elorfin:
| Q             | A
| ------------- | ---
| Bug report?   | no
| Version?      | 11.x (12.x ?)

Not the complete list.

# General
- [x] remove all reference to translation function `tex()` use `trans('key', {}, 'quiz')` instead.
- [ ] move each quiz section (editor, player, etc.) in `#/plugin/exo/resources/quiz` while refactoring the section.
- [ ] rename `enums.js` in `constants.js` and only export a var `constants` which hold all the consts.
- [ ] replace all default exports with named ones.
- [ ] use new reducers syntax.
- [ ] clean selectors (remove duplicates + use reselect)
- [ ] add summary in player and papers (with a quiz param to enable/disable it)
- [x] add progressbar in player (see Path player)
- [ ] make Items fully stand alone (to be able to reuse Editor, Player, Correction) in bank tool.

# Evaluation
- [x] implement `ResourceEvaluation`.

# Overview
- [x] use new `ResourceOverview` core component.

# Player
- [ ] add routing for each step.
- [ ] manage end page route inside player module (for now in resource.jsx)

# Editor
- [x] add routing for each step.
- [x] use standard form reducer and components

# Papers
- [x] use DataListContainer

PR #5529, opened 2019-02-19
[ExoBundle] implements new version

Elorfin:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no

Related to #3872.

- [x] Restore stats. For now it's broken because it requires the full list of papers in the view. Which is no longer available. Statistics API should return computed data (it seems for now it loads all the papers and make calculation in client).


Elorfin:
fixes #5415.
fixes #3922.
fixes #3867.
fixes #5401.

PR #5699, opened 2019-03-28
[PathBundle] Path dashboard + steps deserialization fix

kitan1982:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no


plugin/path/Resources/modules/resources/path/modals/user-progression/components/modal.jsx
```diff
@@ -0,0 +1,79 @@
...
+      </ul>
+    </div>
+
+    <Button
+      className="modal-btn btn btn-primary"
+      type={CALLBACK_BUTTON}
+      primary={false}
+      label={trans('close', {}, 'actions')}
+      callback={() => props.fadeModal()}
+    />
```
Elorfin: No button here. We never add a `.modal-btn` for close (modal can be closed from header or click out).

Also note : 
- `primary={false}` is the default so it's never needed.
- you have a primary class `btn-primary` which is deprecated and is in conflict with your `primary={false}`

plugin/path/Resources/modules/resources/path/modals/user-progression/components/modal.jsx
```diff
@@ -0,0 +1,79 @@
...
+  steps: T.arrayOf(T.shape(StepType.propTypes)).isRequired,
+  stepsProgression: T.object,
+  fadeModal: T.func.isRequired
+}
+
+const UserProgressionModal = connect(
+  (state) => ({
+    stepsProgression: dashboardSelectors.stepsProgression(state)
+  })
+)(UserProgressionModalComponent)
```
Elorfin: You should create a `user-progression/containers/modal` for this.

plugin/path/Resources/modules/resources/path/modals/user-progression/index.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * User progression in path modal.
+ * Displays the progression of an user in the path.
+ */
+
+import {registry} from '#/main/app/modals/registry'
+
+import {UserProgressionModal} from '#/plugin/path/resources/path/modals/user-progression/components/modal'
+
+const MODAL_USER_PROGRESSION = 'MODAL_USER_PROGRESSION'
```
Elorfin: Modal registry is global for app. You should rename it in `MODAL_PATH_USER_PROGRESSION` to avoid later naming conflicts.

Tip : you only have to change the string value :

```
const MODAL_USER_PROGRESSION = 'MODAL_PATH_USER_PROGRESSION'
```

plugin/path/Resources/modules/resources/path/dashboard/components/main.jsx
```diff
@@ -0,0 +1,55 @@
...
+      url: ['innova_path_progressions_fetch', {id: props.path.id}],
+      autoload: true
+    }}
+    primaryAction={(row) => ({
+      type: CALLBACK_BUTTON,
+      callback: () => {
+        props.fetchUserStepsProgression(props.path.id, row.user.id)
+        props.showStepsProgressionModal(row, props.path.steps)
+      }
+    })}
```
Elorfin: It should be a `MODAL_BUTTON` and `props.fetchUserStepsProgression(props.path.id, row.user.id)` should be done in the `onEnter` callback of the modal.

PR #5992, opened 2019-07-11
Fixes user in token + refactors announcement for SPA

kitan1982:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no

### DIFF
--- a/plugin/announcement/Resources/modules/data/sources/announcements.js
+++ b/plugin/announcement/Resources/modules/data/sources/announcements.js
@@ -12,7 +12,7 @@ export default {
       type: URL_BUTTON,
       target: url([ 'claro_resource_show_short', {
         id: announcement.meta.resource.id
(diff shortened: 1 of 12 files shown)


plugin/announcement/Resources/modules/data/sources/announcements.js
```diff
@@ -12,7 +12,7 @@ export default {
       type: URL_BUTTON,
       target: url([ 'claro_resource_show_short', {
         id: announcement.meta.resource.id
-      }]) + `#/${announcement.id}`
+      }]) + `#/desktop/resource_manager/${announcement.meta.resource.id}/${announcement.id}`
```
Elorfin: ça doit être un `LINK_BUTTON` et du coup le target c'est : `/desktop/resource_manager/${announcement.meta.resource.id}/${announcement.id}`.

PR #6827, opened 2020-09-12
Fix framework-extra-bundle + doctrine/common deprecations

chalasr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Fixed issues | continuation of #6746 

- Moves from `doctrine/common` persistence API (deprecated) to `doctrine/persistence`
- Migrates routing configuration from `sensio/framework-extra-bundle` to `symfony/routing` (`@Route` annotation only)
- Fixes `symfony/config`'s `TreeBuilder` deprecations
- Adds `composer.lock` to `.gitignore`


.gitignore
```diff
@@ -3,3 +3,4 @@ node_modules/*
 package-lock.json
 .php_cs.cache
 .phpunit.result.cache
+composer.lock
```
Elorfin, line 6: I think this should be in : https://github.com/claroline/Claroline/blob/13.x/.gitignore.

chalasr: Would it harm to have it here as well? Working on `claroline/distribution` library outside of the `claroline/claroline` project should be possible, and you don't want to commit the lock file when doing so (given it's untracked by now).
Also as a rule of thumb, libraries should not commit their lock file while deployed projects should (https://getcomposer.org/doc/04-schema.md#type).

composer.json
```diff
@@ -39,16 +39,15 @@
...
         "stefk/jval": "dev-master",
         "stof/doctrine-extensions-bundle": "1.3.0",
-        "symfony/monolog-bundle": "3.2.0",
+        "symfony/monolog-bundle": "^3.5",
         "symfony/asset": "4.4.*",
         "symfony/browser-kit": "4.4.*",
         "symfony/cache": "4.4.*",
         "symfony/config": "4.4.*",
         "symfony/console": "4.4.*",
-        "symfony/debug": "4.4.*",
```
chalasr, line 51: Deprecated in favor of symfony/error-handler.

integration/big-blue-button/Controller/BBBController.php
```diff
@@ -121,7 +122,7 @@ public function moderatorMeetingsListAction()
     }
 
     /**
-     * @EXT\Route("/{id}/meeting", name="apiv2_bbb_meeting_create")
+     * @Route("/{id}/meeting", name="apiv2_bbb_meeting_create")
      * @EXT\Method("POST")
```
Elorfin, line 126: Why not loading this from symfony components too ?

chalasr: The `@Method` annotation disappears, we have to migrate to an attribute of the `@Route` one (i.e. `@Route(methods={"GET"}`. That's for my next PR :)

Elorfin (approved):

PR #6838, opened 2020-09-17
switch annotation from sensio/framework-extra-bundle to symfony/routing

GrimpEreau:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Fixed issues |

### DIFF
--- a/main/core/Controller/APINew/Template/TemplateController.php
+++ b/main/core/Controller/APINew/Template/TemplateController.php
@@ -133,14 +133,14 @@ class TemplateController extends AbstractCrudController
     /**
      * @Route(
      *     "{id}/default",
-     *     name="apiv2_template_default_define"
+     *     name="apiv2_template_default_define",
+     *     methods={"PUT"}
      * )
      * @EXT\ParamConverter(
      *     "template",
      *     class="ClarolineCoreBundle:Template\Template",
      *     options={"mapping": {"id": "uuid"}}
      * )
-     * @EXT\Method("PUT")
      *
      * @param Template $template
      *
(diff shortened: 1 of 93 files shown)


Elorfin (approved):

chalasr (changes requested):
This seems to break tests. No merge please, looking at it.

main/core/Controller/APINew/Template/TemplateController.php
```diff
@@ -138,9 +138,9 @@ public function templatesByLangDeleteBulkAction(Request $request)
      * @EXT\ParamConverter(
      *     "template",
      *     class="ClarolineCoreBundle:Template\Template",
-     *     options={"mapping": {"id": "uuid"}}
+     *     options={"mapping": {"id": "uuid"}},
+     *     methods={"PUT"}
```
chalasr: This is the culprit! It's not a Route but a ParamConverter :)

main/core/Controller/APINew/User/LocationController.php
```diff
@@ -35,7 +35,7 @@ public function getName()
     use HasOrganizationsTrait;
 
     /**
-     * @Route("/{id}/geolocate", name="apiv2_location_geolocate", methods={"GET"})
+     * @Route("/{id}/geolocate", name="apiv2_location_geolocate@", methods={"GET"})
```
Elorfin: I think this is a typo.

chalasr: Good catch, fixed. thanks

chalasr (approved):

PR #6885, opened 2020-10-09
Don't execute updaters more than once

chalasr:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Fixed issues | -

- [x] Store updaters execution in DB
- [x] By default, ensure they are never executed twice
- [x] Allow using `claroline:update --force` to execute any updater no matter if they have been already
- [x] Add `claroline:updater:execute [UPDATER CLASS]` to manually run a given updater.
- [x] Add marker interface for updaters that cannot be executed more than once, regardless of the `--force` flag

### DIFF
--- a/main/core/Resources/config/services/command.yml
+++ b/main/core/Resources/config/services/command.yml
@@ -114,6 +114,14 @@ services:
         tags:
             - { name: 'console.command', command: 'claroline:update' }
 
+    Claroline\CoreBundle\Command\Dev\ExecuteUpdaterCommand:
+        arguments:
+            - '@Claroline\InstallationBundle\Repository\UpdaterExecutionRepository'
+            - !tagged_locator
+                  tag: 'claroline.platform.updater'
+        tags:
+            - { name: 'console.command', command: 'claroline:updater:execute' }
+
     Claroline\CoreBundle\Command\Dev\RefreshCommand:
         arguments:
             - '@Claroline\CoreBundle\Library\Installation\Refresher'
--- a/plugin/cursus/Installation/AdditionalInstaller.php
+++ b/plugin/cursus/Installation/AdditionalInstaller.php
@@ -2,25 +2,15 @@
 
 namespace Claroline\CursusBundle\Installation;
 
+use Claroline\CursusBundle\Installation\Updater\Updater130000;
 use Claroline\InstallationBundle\Additional\AdditionalInstaller as BaseInstaller;
 
 class AdditionalInstaller extends BaseInstaller
 {
-    public function preUpdate($currentVersion, $targetVersion)
+    public static function getUpdaters(): array
     {
-        if (version_compare($currentVersion, '13.0.0', '<')) {
-            $updater = new Updater\Updater130000($this->container, $this->logger);
-            $updater->setLogger($this->logger);
-            $updater->preUpdate();
-        }
-    }
-
-    public function postUpdate($currentVersion, $targetVersion)
-    {
-        if (version_compare($currentVersion, '13.0.0', '<')) {
-            $updater = new Updater\Updater130000($this->container, $this->logger);
-            $updater->setLogger($this->logger);
-            $updater->postUpdate();
-        }
+        return [
+            '13.0.0' => Updater130000::class,
+        ];
     }
 }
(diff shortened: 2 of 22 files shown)


plugin/cursus/Installation/AdditionalInstaller.php
```diff
@@ -2,25 +2,37 @@
...
+
+            if (!$this->shouldReplayUpdaters() && $alreadyExecuted) {
+                return;
+            }
+
+            $updater = $this->container->get(Updater130000::class);
             $updater->postUpdate();
+
+            if (!$alreadyExecuted) {
+                $this->markUpdaterAsExecuted(Updater\Updater130000::class);
```
Elorfin: You don't need the `Updater\` part.

plugin/cursus/Installation/AdditionalInstaller.php
```diff
@@ -2,25 +2,37 @@
...
 {
     public function preUpdate($currentVersion, $targetVersion)
     {
         if (version_compare($currentVersion, '13.0.0', '<')) {
-            $updater = new Updater\Updater130000($this->container, $this->logger);
-            $updater->setLogger($this->logger);
+            if (!$this->shouldReplayUpdaters() && $this->isUpdaterAlreadyExecuted(Updater130000::class)) {
+                return;
+            }
+            $updater = $this->container->get(Updater130000::class);
```
Elorfin: IMO Updaters should not be public.
We have reset all Migrations/Updaters for 13.x, all is clean and shiny but this things grow really fast and container will contain hundred of this services which can never be used elsewhere.

I'm OK to use standard DI system instead of manual `new` + container injection but at least we need to rely on tagged services and an intermediate provider service class to collect/retrieve them and make them private.

Elorfin (commented):
Thank you for the PR.

- The command to manually execute Updaters should also allow to execute `preInstall` / `postInstall`.
- We should find a way to automatically execute all the additional code in `AdditionalInstaller` (mark as executed + checks).
It will be painful to write it for each updater, and the class will quickly become unreadable as we add more updaters.

I'm not sure about this one, but I think we need to dissociate `pre` / `post` action. For now, an Updater is marked as executed if both pre and post have been executed. And if we want to replay it we will not be able to replay only post or only pre. To be discussed at the next meeting.

plugin/cursus/Installation/AdditionalInstaller.php
chalasr: Fully agreed. Patch updated accordingly

plugin/cursus/Installation/AdditionalInstaller.php
```diff
@@ -2,25 +2,15 @@
...
-
-    public function postUpdate($currentVersion, $targetVersion)
-    {
-        if (version_compare($currentVersion, '13.0.0', '<')) {
-            $updater = new Updater\Updater130000($this->container, $this->logger);
-            $updater->setLogger($this->logger);
-            $updater->postUpdate();
-        }
+        return [
+            '13.0.1' => Updater130000::class,
```
chalasr: for local testing, should be `13.0.0`

Elorfin (approved):

main/core/Resources/config/services/command.yml
```diff
@@ -114,6 +114,14 @@ services:
         tags:
             - { name: 'console.command', command: 'claroline:update' }
 
+    Claroline\CoreBundle\Command\Dev\ExecuteUpdaterCommand:
+        arguments:
+            - '@Claroline\InstallationBundle\Repository\UpdaterExecutionRepository'
+            - !tagged_locator
+                  tag: 'claroline.platform.updater'
```
chalasr, line 121: As promised:
https://symfony.com/doc/current/service_container/service_subscribers_locators.html
https://symfony.com/doc/current/service_container/service_subscribers_locators.html#indexing-the-collection-of-services

Elorfin: Thank you 👍

chalasr:
Not sure we want to fix the remaining phpmd failure. 
@Elorfin Can you merge this one please?

PR #6904, opened 2020-10-16
fixes templating service

Elorfin:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no

- Fixes access to `templating` service from the container.
- Updates templates shortnames. It seems like the synthax we were using is not supported (`Template reference "ClarolineOpenBadgeBundle::javascripts.html.twig" not found, did you mean "@ClarolineOpenBadge/javascripts.html.twig"?`).

@Zowac @chalasr FYI.


chalasr (approved):
Oops! Thanks @Elorfin

PR #6924, opened 2020-10-26
Cleanup session wiring

chalasr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Fixed issues | -

The `session.handler_id` configuration processing changes when its value is an environment variable (it expects a DSN e.g. `redis://localhost` or `file://some_dir`). As we don't need to change the session handler at runtime, we can switch to a regular parameter for the session handler id.

The session configuration is located in `platform_options.json`.

This also removes code related to a previous custom session_handler implementation (`pdo|claro_pdo`) as it short-circuits the new implementation and is not needed anymore.


Elorfin (approved):
Seems OK to me. Thank you.

chalasr:
@LaurentGruber Here are the new defaults for session storage (platform options):

```json
"session": {
    "storage_type": "file",
    "redis_host": "localhost",
    "redis_port": 6379,
    "redis_password": "",
}
```

By default, the file storage type is used (`storage_type` can be set to `file` or `native` for that, defaults to `file`).
To enable the redis session storage, set the `storage_type` option to `redis` and change the redis related options (_host, _port, _password), e.g:

```json
"session": {
    "storage_type": "redis",
    "redis_host": "localhost",
    "redis_port": 6379,
    "redis_password": "dummy",
}
```

PR #6952, opened 2020-11-12
Unlock composer 2 compat by removing automagic bundle detection

chalasr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | for the better, yes
| Fixed issues | -

Adding a distributed `bundles.ini-dist` file to maintain in this repository, which is copied on composer install/update *if it does not exist*.

Given the removed bundle detection was built on the composer 1 API, we can now remove the `composer self-update --1` trick from the travis build once this is merged.

Durations for `composer update` on `claroline/claroline` locally:
with composer 1:  136.87s
with composer 2: 63.93s


chalasr:
Don't merge yet :)

Elorfin (commented):
We may need to write an updater to remove third party bundles from `bundles.ini` if symfony returns an error when we try to load a bundle more than once.

chalasr:
@Elorfin Third party bundles aren't listed in the final bundles.ini anymore, as they are now an internal implementation detail of claroline plugins. 
As discussed, if one want to register a custom bundle, one have to write a claroline plugin that registers it via `getRequiredThirdPartyBundles()`.
I just pushed a change that ensures uniqueness for 3rd party bundles (indexing bundles by class name in the kernel registration process does the trick).

Please let me know  if you are talking about something else or if it's ok for you

Elorfin:
I was thinking of the old platforms : third parties are present in their `bundles.ini` and will also be loaded from code in the new version.
With your changes, no need to worry about it anymore ;). Thank you.

Elorfin (approved):

PR #6957, opened 2020-11-16
Fix deprecations:

Zowac:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no

Fix deprecations : 
 - Controller is deprecated since Symfony 4.2, use AbstractController instead
 - Replacing "interface AdvancedUserInterface" wich is deprecated with a custom user checker

Sponsored by @coopTilleuls


Elorfin (approved):
Thank you.

PR #6961, opened 2020-11-23
Switch from swiftmailer to mailer

Zowac:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Fixed issues | -

Reduces the vendor list by replacing swiftmailer by symfony/mailer which has a better, moder and well maintained API.


main/core/Library/Mailing/Client/SymfonyMailer.php
```diff
@@ -0,0 +1,40 @@
...
+            $email->attachFromPath(
+                $attachment['path']
+            );
+        }
+
+        try {
+            $this->mailer->send($message);
+
+            return true;
+        } catch (\Exception $e) {
```
chalasr: Only exceptions from the symfony/mailer component need to be caught (https://github.com/symfony/symfony/blob/5.x/src/Symfony/Component/Mailer/Exception/ExceptionInterface.php)

main/core/Library/Mailing/Client/SymfonyMailer.php
```diff
@@ -0,0 +1,40 @@
...
+            ->from($message->getAttribute('from'))
+            ->replyTo($message->getAttribute('reply_to'))
+            ->html($message->getAttribute('body'))
+            ->bcc(...$message->getAttribute('bcc'))
+            ->to(...$message->getAttribute('to'));
+
+        foreach ($message->getAttribute('attachments') as $attachment) {
+            $email->attachFromPath(
+                $attachment['path']
+            );
```
chalasr: nitpick: should be on one line :)

main/core/Library/Mailing/TransportFactory.php
```diff
@@ -12,72 +12,73 @@
 namespace Claroline\CoreBundle\Library\Mailing;
 
 use Claroline\CoreBundle\Library\Configuration\PlatformConfigurationHandler;
+use Monolog\Logger;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
```
chalasr:
```suggestion
use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
```

main/core/Library/Mailing/TransportFactory.php
```diff
@@ -12,72 +12,73 @@
 namespace Claroline\CoreBundle\Library\Mailing;
 
 use Claroline\CoreBundle\Library\Configuration\PlatformConfigurationHandler;
+use Monolog\Logger;
```
chalasr: Better use high-level abstractions.

```suggestion
use Psr\Log\LoggerInterface;
```

main/core/Library/Mailing/TransportFactory.php
```diff
@@ -12,72 +12,73 @@
...
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\Mailer\Bridge\Google\Transport\GmailSmtpTransport;
+use Symfony\Component\Mailer\Transport\SendmailTransport;
+use Symfony\Component\Mailer\Transport\Smtp\EsmtpTransport;
 
 class TransportFactory
 {
+    private const COMMAND = '/usr/sbin/sendmail -bs';
+    private const TIME_OUT = 30;
```
chalasr: in one word: `TIMEOUT` :)

main/core/Library/Mailing/TransportFactory.php
```diff
@@ -12,72 +12,73 @@
...
             $transport->setPassword($this->configHandler->getParameter('mailer_password'));
 
             return $transport;
         }
 
-        //default
-        return $this->getBaseSmtpTransport();
-    }
+        // Default smtp
+        $encryption = null;
```
chalasr:
```suggestion
        $encryption = null; // Lets the transport choose the best value based on the platform.
```

main/core/Library/Mailing/TransportFactory.php
```diff
@@ -12,72 +12,73 @@
...
+        if (!empty($this->configHandler->getParameter('mailer_encryption')) && 'none' !== $this->configHandler->getParameter('mailer_encryption')) {
```
chalasr: As briefly discussed, `none` should mean "no encryption"  (i.e. `false`). The check should be updated accordingly

main/core/Tests/Unit/Library/Mailing/Client/MailerTest.php
```diff
@@ -0,0 +1,65 @@
...
+
+namespace Claroline\CoreBundle\Tests\Unit\Library\Mailing\Client;
+
+use Claroline\CoreBundle\Library\Mailing\Client\SymfonyMailer;
+use Claroline\CoreBundle\Library\Mailing\Message;
+use Claroline\CoreBundle\Library\Testing\MockeryTestCase;
+use Symfony\Component\Mailer\MailerInterface;
+use Symfony\Component\Mime\Email;
+
+class MailerTest extends MockeryTestCase
```
chalasr:
```suggestion
class SymfonyMailerTest extends MockeryTestCase
```

main/core/Tests/Unit/Library/Mailing/Client/MailerTest.php
```diff
@@ -0,0 +1,65 @@
...
+{
+    public function testEmailWithMessageData()
```
chalasr: This test does not bring much value as it is more about testing the `Email` class which is already tested on symfony side. I suggest to remove it

main/core/Tests/Unit/Library/Mailing/Client/MailerTest.php
```diff
@@ -0,0 +1,65 @@
...
+        $this->assertSame('reply_to@claroline.com', $message->getAttribute('reply_to'));
+        $this->assertSame('<p>Hello</p>', $message->getAttribute('body'));
+        $this->assertSame('bcc1@claroline.com', $message->getAttribute('bcc')[0]);
+        $this->assertSame('bcc2@claroline.com', $message->getAttribute('bcc')[1]);
+        $this->assertCount(2, $message->getAttribute('bcc'));
+        $this->assertSame('to@claroline.com', $message->getAttribute('to')[0]);
+        $this->assertTrue($symfonyMailer->send($message));
+    }
+
+    public function getMessageData()
```
chalasr: The body of this method can be directly inlined in `testSend()` above, given it is the only relevant test

main/core/Library/Mailing/Client/SymfonyMailer.php
chalasr: Also I suggest to log that an error occurred while trying to send the mail, with the exception passed in the log context.

main/core/DependencyInjection/Compiler/DynamicConfigPass.php
```diff
@@ -34,8 +34,8 @@ public function process(ContainerBuilder $container)
             new Reference('Claroline\CoreBundle\Library\Mailing\TransportFactory'),
```
chalasr, line 36: `Swift_Transport` does not exist anymore (line 32) :) 
The class should be `TransportInterface` from symfony/mailer (because it is a factory, we can use a broader type, no need to set concrete class).

main/core/DependencyInjection/Compiler/DynamicConfigPass.php
```diff
@@ -34,8 +34,8 @@ public function process(ContainerBuilder $container)
...
             'getTransport', ]
         );
-        $container->removeDefinition('swiftmailer.mailer.default.transport');
-        $container->setDefinition('swiftmailer.mailer.default.transport', $transport);
+        $container->removeDefinition('mailer.default_transport');
```
chalasr, line 39: That `mailer.default_transport` should be configured somewhere :)
What is missing here is to replace the suggested config for `swiftmailer` by a minimal config for `mailer` (which will make the framework create the `@mailer` and `@mailer.default_transport` services).

chalasr (changes requested):
Great work so far!

main/core/DependencyInjection/Compiler/DynamicConfigPass.php
```diff
@@ -29,7 +29,7 @@ public function process(ContainerBuilder $container)
     {
         //mailing
         $transport = new Definition();
-        $transport->setClass('Swift_Transport');
+        $transport->setClass('TransportInterface');
```
chalasr: `TransportInterface::class` (with import for ` Symfony\Component\Mailer\Transport\TransportInterface`)

Zowac: Oops, fixed, thx :)

main/core/Library/Mailing/Client/SymfonyMailer.php
```diff
@@ -0,0 +1,46 @@
...
+        foreach ($message->getAttribute('attachments') as $attachment) {
+            $email->attachFromPath($attachment['path']);
+        }
+
+        try {
+            $this->mailer->send($message);
+
+            return true;
+        } catch (ErrorInterface $e) {
+            $this->logger->error($e->getMessage());
```
chalasr:
```suggestion
            $this->logger->error($e->getMessage(), ['exception' => $e]);
```

main/core/Library/Mailing/TransportFactory.php
```diff
@@ -12,72 +12,73 @@
...
+            $this->configHandler->getParameter('mailer_port'),
+            $encryption,
+            $this->eventDispatcher,
+            $this->logger
         );
+        $transport->setUsername($this->configHandler->getParameter('mailer_username'));
+        $transport->setPassword($this->configHandler->getParameter('mailer_password'));
+        // should probably be configurable too
+        $transport->getStream()->setTimeout(self::TIMEOUT);
+        $transport->setSourceIp(null);
```
chalasr: Does this method actually exist?

main/core/Library/Mailing/Client/SymfonyMailer.php
Zowac: Done, thx :)

main/core/Library/Mailing/TransportFactory.php
Zowac: Yes, but i forget `getStream` before and after cheking the method doesn't support null value, i removed this line. Thanks you

main/core/DependencyInjection/Compiler/DynamicConfigPass.php
Zowac: Done, thx :)

main/core/Library/Mailing/Client/SymfonyMailer.php
Zowac: Done, thx :)

main/core/Library/Mailing/Client/SymfonyMailer.php
Zowac: Done, thx :)

main/core/Library/Mailing/Client/SymfonyMailer.php
```diff
@@ -0,0 +1,56 @@
...
+
+        foreach ($message->getAttribute('attachments') as $attachment) {
+            $email->attachFromPath($attachment['path']);
+        }
+
+        try {
+            $this->mailer->send($message);
+
+            return true;
+        } catch (ExceptionInterface $e) {
```
chalasr: I just realized that the `Mailer` class is responsible for catching mailing-related errors and logging them to a specific file. Let's remove the try-catch

chalasr (changes requested):

main/core/Library/Mailing/Client/SymfonyMailer.php
Zowac: Indeed, thank you :)

chalasr (approved):
👍  Tested locally, all good (both smtp and gmail).
Thanks @Zowac

Zowac:
Thanks for your help @chalasr

main/core/Library/Mailing/TransportFactory.php
```diff
@@ -12,72 +12,70 @@
...
-                        new \Swift_Transport_Esmtp_Auth_CramMd5Authenticator(),
-                        new \Swift_Transport_Esmtp_Auth_LoginAuthenticator(),
-                        new \Swift_Transport_Esmtp_Auth_PlainAuthenticator(),
-                    ]
-                ),
-            ],
-            new \Swift_Events_SimpleEventDispatcher()
+        $transport = new EsmtpTransport(
+            $this->configHandler->getParameter('mailer_host'),
+            $this->configHandler->getParameter('mailer_port'),
```
Elorfin: It is `null` when using postal. It breaks the container compilation so we can't migrate it and it should be handled by code.

```
$this->configHandler->getParameter('mailer_port') ?? 0,
```

main/core/Library/Mailing/TransportFactory.php
```diff
@@ -12,72 +12,70 @@
...
+            $encryption,
+            $this->eventDispatcher,
+            $this->logger
         );
+        $transport->setUsername($this->configHandler->getParameter('mailer_username'));
+        $transport->setPassword($this->configHandler->getParameter('mailer_password'));
```
Elorfin: Same thing with `mailer_username` and `mailer_transport`, `null` values are not allowed by `EsmtpTransport`.

Elorfin (changes requested):
Thank you for the work.
It just needs little changes to make `postal` work on our existing platforms.

Zowac:
@Elorfin 

The changes are done, thank you for your feedback :)

Elorfin (approved):

PR #6973, opened 2020-11-30
Fix install on composer 2

chalasr:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Fixed issues | -

Mistakenly introduced in #6952


Elorfin (approved):
Thank you. I've just rewritten it to avoid using composer.

chalasr:
Even better! 👍thanks

